PR #51, opened 2017-10-29
compress static files

bobzoller:
@paulcsmith I took a stab at luckyframework/web#154.

My Ruby is rusty and I'm new to Crystal, so stylistic feedback is welcomed.  Other questions:
- feels like this wants to be `LuckyWeb::StaticFileHandler` and live in the web repo?
- thoughts about my re-use of Crystal's standard `HTTP::CompressHandler` rather than copy pasta?
- will add a mime-type filter for compression (IIRC it's best practice not to (re)compress images, for example), but wanted to make sure I'm on the right track first.

### DIFF
--- a/src/web_app_skeleton/src/handlers/static_file_handler.cr
+++ b/src/web_app_skeleton/src/handlers/static_file_handler.cr
@@ -1,4 +1,17 @@
-class HTTP::StaticFileHandler
+require "http"
+
+class StaticFileHandler < HTTP::StaticFileHandler
+  @compress_handler : HTTP::Handler = HTTP::CompressHandler.new
+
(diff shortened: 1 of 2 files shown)


paulcsmith:
Sorry, I didn't see this so I added some comments here: https://github.com/luckyframework/web/issues/154#issuecomment-340459207

I'll close this so we can continue on the web repo. But yes the handler should be in the `web` repo. Not sure why I ever decided to put the original implementation in this repo :)

paulcsmith:
May also be worth looking at adding better mime type support similar to this: https://github.com/amberframework/amber/blob/master/src/amber/router/pipe/static.cr#L76

Issue #47, opened 2017-10-27
`lucky init` fails when directory already exists


ianagne:
## Steps to repeat:
1) create a new directory (e.g. `lucky_charms/` )
2) run `lucky init lucky_charms`

## Error output:
```
file or directory lucky_charms already exists
Adding deps to shards.yml
Error opening file 'shard.yml' with mode 'r': No such file or directory (Errno)
0x104d4be80: *raise<Errno>:NoReturn at ??
0x104d58b34: *File::new<String, String, Int32, Nil, Nil>:File at ??
0x104d74def: *LuckyCli::Generators::Web@LuckyCli::GeneratorHelpers#append_text<String, String>:Nil a0x104d74def: *LuckyCli::Generators::Web@LuckyCli::GeneratorHelpers#append_text<String, String>:Nil at ??
Unhandled exception in spawn:
End of file reached (IO::EOFError)
0x104d5a480: *raise<IO::EOFError>:NoReturn at ??
0x104d699d4: ~procProc(Nil)@/usr/local/Cellar/crystal-lang/0.23.1_3/src/event/signal_handler.cr:93 at ??
0x104d71af6: *LuckyCli::Generators::Web#run:Nil at ??
0x104d49f65: __crystal_main at ??
0x104d6a7f5: main at ??
```
## Desired behavior

I expected the project to initialize since the directory was empty.  Thoughts?  I'd love to help out with this.

paulcsmith:
@ianagne Good catch! I think it should not initialize if the project is empty, but should raise a prettier error like

```
There is already a directory called {project_name}. Please delete it and try again, or name your project something different.
```

In the future we could ask "this directory is empty, would you like to initialize the project in it?", but for now I think it's best to keep things simple.

If you (or someone else) wants to take a crack at it, that'd be great. You would probably want to start here: https://github.com/luckyframework/cli/blob/master/src/generators/web.cr

That class generates a new lucky web project. In the `#run` method you would probably start off with a check like `ensure_directory_does_not_exist` or something along those lines.

coderhs:
@paulcsmith I would like to have go with this issue. 😄 👍

paulcsmith:
@coderhs That'd be great! Thank you

coderhs:
I just noticed an interesting point. let me know if we want to handle that.

If the folder that exist is a crystal project, then it won't fail but add our code to that project. 

```
crystal init app lucky_charms
lucky init lucky_charms
```

[...]

PR #68, opened 2017-11-10
Fix: #47; ensure directory with same name does not exists

coderhs:
This PR is in reference to issue #47

### DIFF
--- a/spec/integration/init_web_spec.cr
+++ b/spec/integration/init_web_spec.cr
@@ -2,28 +2,34 @@ require "../spec_helper"
 
 describe "Initializing a new web project" do
   it "creates the project successfully" do
-    puts "Running integration spec. This might take awhile...".colorize(:yellow)
-    should_run_successfully "rm -rf ./test-project"
-    should_run_successfully "crystal src/lucky.cr init test-project"
-    FileUtils.cd "test-project" do
-      should_run_successfully "crystal build src/server.cr"
-      should_run_successfully "crystal build src/test_project.cr"
-      should_run_successfully "crystal src/app.cr"
+    begin
+      puts "Running integration spec. This might take awhile...".colorize(:yellow)
+      should_run_successfully "rm -rf ./test-project"
+      should_run_successfully "crystal src/lucky.cr init test-project"
+      FileUtils.cd "test-project" do
+        should_run_successfully "crystal build src/server.cr"
+        should_run_successfully "crystal build src/test_project.cr"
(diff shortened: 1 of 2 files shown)


spec/integration/init_web_spec.cr
```diff
@@ -10,6 +10,20 @@ describe "Initializing a new web project" do
       should_run_successfully "crystal build src/test_project.cr"
       should_run_successfully "crystal src/app.cr"
     end
+    FileUtils.rm_r "test-project"
```
paulcsmith: I think this should go in a begin/ensure so that it's always deleted at the end. Something like this:

```crystal
begin do
  # the test code
ensure
  FIleUtils.rm_r "test-project"
end
```

spec/integration/init_web_spec.cr
```diff
@@ -10,6 +10,20 @@ describe "Initializing a new web project" do
...
+  it "does not create project if directory with same name already exist" do
+    FileUtils.mkdir "test-project"
+    output = IO::Memory.new
+    Process.run(
+      "crystal src/lucky.cr init test-project",
+      output: output,
+      shell: true
+    )
+    FileUtils.rm_r "test-project"
+    message = "Folder with test-project already exists, please use a different name"
```
paulcsmith: Maybe "Folder *named*" would make more sense here

spec/integration/init_web_spec.cr
```diff
@@ -10,6 +10,20 @@ describe "Initializing a new web project" do
...
+  end
+
+  it "does not create project if directory with same name already exist" do
+    FileUtils.mkdir "test-project"
+    output = IO::Memory.new
+    Process.run(
+      "crystal src/lucky.cr init test-project",
+      output: output,
+      shell: true
+    )
```
paulcsmith: Nice!

spec/integration/init_web_spec.cr
```diff
@@ -10,6 +10,20 @@ describe "Initializing a new web project" do
...
+    FileUtils.rm_r "test-project"
```
paulcsmith: I think this should also be done with begin/ensure so it's always deleted

spec/integration/init_web_spec.cr
coderhs: Yes, that would be the best way to do it. I will update the code.

paulcsmith:
This is really nice. Less errors = good 👍 Thanks!

Issue #63, opened 2017-11-06
If `yarn` isn't installed let people know


paulcsmith:
Maybe in bin/compile_assets? https://github.com/luckyframework/cli/blob/master/src/web_app_skeleton/bin/compile_assets

See https://gitter.im/luckyframework/Lobby?at=59ff95f7976e63937e2f6fbf

paulcsmith:
Closed by https://github.com/luckyframework/cli/commit/066bc12a04055bbd0600ca22e69ad51d62c87b45

PR #69, opened 2017-11-10
Add warning if yarn or brunch not installed

ianagne:
Users will be unable to init a lucky application if they do not have
yarn or brunch installed.  There are existing warnings in the
documentation that these are dependencies, but the extra heads up is a
friendlier interaction when the init script fails to complete.

See issue [#63](https://github.com/luckyframework/cli/issues/63)

### DIFF
--- a/src/web_app_skeleton/bin/setup
+++ b/src/web_app_skeleton/bin/setup
@@ -9,12 +9,6 @@ if ! command -v yarn > /dev/null; then
   exit 1
 fi
 
-if ! command -v brunch > /dev/null; then
-  printf 'Brunch is not installed.\n'
-  printf 'Run "npm install -g brunch" to install brunch.\n'
-  exit 1
-fi
-
 echo "\n▸ Installing node dependencies"
 yarn install
 
--- a/src/web_app_skeleton/bin/setup
+++ b/src/web_app_skeleton/bin/setup
@@ -3,6 +3,18 @@
 # Exit if any subcommand fails
 set -e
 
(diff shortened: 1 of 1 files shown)


ianagne:
`yarn install` is called before the compile_assets script is run, so I opted to put the check in the setup script instead.

src/web_app_skeleton/bin/setup
```diff
@@ -3,6 +3,18 @@
 # Exit if any subcommand fails
 set -e
 
+if ! command -v yarn > /dev/null; then
+  printf 'Yarn is not installed.\n'
+  printf 'See https://yarnpkg.com/lang/en/docs/install/ for install instructions.\n'
+  exit 1
+fi
```
paulcsmith: Very nice! Thank you so much

src/web_app_skeleton/bin/setup
```diff
@@ -3,6 +3,18 @@
...
+
+if ! command -v brunch > /dev/null; then
+  printf 'Brunch is not installed.\n'
+  printf 'Run "npm install -g brunch" to install brunch.\n'
+  exit 1
+fi
```
paulcsmith: You don't need this because if you have yarn then `yarn install` will install it. `bin/compile_assets` uses the brunch version installed in the project directory, so you don't need to install it globally

ianagne: 👍

ianagne:
Sounds like a plan.  I'll add a check for `heroku` as well.

PR #80, opened 2017-11-15
Use render method instead of block in app skeleton

### LINKED ISSUE #77: Generate layout with regular `render` method
https://github.com/luckyframework/web/pull/213

arr-dev:
Closes #77

### DIFF
--- a/src/web_app_skeleton/src/pages/main_layout.cr
+++ b/src/web_app_skeleton/src/pages/main_layout.cr
@@ -11,7 +11,7 @@ abstract class MainLayout
 
   abstract def inner
 
-  render do
+  def render
     html_doctype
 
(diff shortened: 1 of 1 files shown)


paulcsmith:
Thanks :D

arr-dev:
Noticed that specs `spec/lucky_cli/runner_spec.cr` specs fail due to task from `spec/lucky_cli/task_spec.cr` leak into runner_spec.
Probably the best way to fix it is to define all spec tasks in one support file?

paulcsmith:
@arr-dev I just rebased against master and it looks like all specs pass. Can you get the latest from master and try again? 

Or if it still doesn't work can you open an issue with the output?

Issue #23, opened 2017-06-30
Install yarn silently


paulcsmith:
So much noise when initially setting up a project :S Consider using `--silent` https://github.com/yarnpkg/yarn/issues/788

paulcsmith:
Closed by https://github.com/luckyframework/lucky_cli/pull/93#merged-event

Issue #84, opened 2017-11-15
Update package.json to fix warnings


paulcsmith:
```
▸ Installing node dependencies
  yarn install v1.2.1
warning ../../../package.json: No license field
  info No lockfile found.
  [1/4] Resolving packages...
warning babel-brunch > babel-preset-latest@6.24.1: We're super 😸  excited that you're trying to use ES2017+ syntax, but instead of making more yearly presets 😭 , Babel now has a better preset that we recommend you use instead: npm install babel-preset-env --save-dev. preset-env without options will compile ES2015+ down to ES5 just like using all the presets together and thus is more future proof. It also allows you to target specific browsers so that Babel can do less work and you can ship native ES2015+ to user 😎 ! We are also in the process of releasing v7, so please give http://babeljs.io/blog/2017/09/12/planning-for-7.0 a read and help test it out in beta! Thanks so much for using Babel 🙏, please give us a follow on Twitter @babeljs for news on Babel, join slack.babeljs.io for discussion/development and help support the project at opencollective.com/babel
warning babel-brunch > babel-preset-latest > babel-preset-es2017@6.24.1: 🙌  Thanks for using Babel: we recommend using babel-preset-env now: please read babeljs.io/env to update!
warning babel-brunch > babel-preset-latest > babel-preset-es2016@6.24.1: 🙌  Thanks for using Babel: we recommend using babel-preset-env now: please read babeljs.io/env to update!
warning babel-brunch > babel-preset-latest > babel-preset-es2015@6.24.1: 🙌  Thanks for using Babel: we recommend using babel-preset-env now: please read babeljs.io/env to update!
warning browser-sync > browser-sync-ui > weinre > express@2.5.11: express 2.x series is deprecated
warning browser-sync > browser-sync-ui > weinre > express > connect@1.9.2: connect 1.x series is deprecated
```

* Needs a license field. Not sure what the default should be. Maybe a "None" or "Unknown"? See if NPM has any defaults that they use when the license is not specified
* Fix all the other warnings

You can test this out and see the output by running `crystal spec spec/integration`

PR #93, opened 2017-11-26
Install yarn dependencies silently

ianagne:
Yarn outputs significant noise while installing node dependencies.
Outputting excessively to STDOUT during normal operation can hide
relevant error messages from users.  Install yarn with silent flag so
only errors are outputted when running `yarn install`

See issue #23

### DIFF
--- a/src/web_app_skeleton/bin/setup
+++ b/src/web_app_skeleton/bin/setup
@@ -23,6 +23,7 @@ indent() {
 
 echo "\n▸ Installing node dependencies"
 yarn install --silent | indent
+echo "\n▸ Done installing node dependencies"
 
 echo "\n▸ Compiling assets"
 bin/compile_assets build | indent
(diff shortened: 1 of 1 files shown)


ianagne:
The issue just said that you were considering adding the silent flag.  The PR is here if you want it 😄

src/web_app_skeleton/bin/setup
```diff
@@ -22,7 +22,7 @@ indent() {
 }
 
 echo "\n▸ Installing node dependencies"
-yarn install | indent
+yarn install --silent | indent
```
paulcsmith: Could you also echo "Done installing node dependencies | indent" so it isn't totally silent? Once the package.json warnings are fixed (#84) this will be extra important because it'll look like nothing happened if we don't print a success message:

ianagne: 👍

paulcsmith:
Thanks! :D

PR #95, opened 2017-11-29
Fix typo /succeful/successful/

hanneskaeufler:
🤖


paulcsmith:
I guess that matcher was unusucceful :P

paulcsmith:
Thanks for catching that 👍

Issue #100, opened 2017-12-02
Set LUCKY_ENV to spec helper


edwardloveall:
Without this, it's pretty easy to erase your dev database when running tests. I thought it would be as simple as adding this to the top of the generated spec_helper.cr:

```crystal
ENV["LUCKY_ENV"] = "test"
```

but I don't see a generated spec_helper in the app skeleton.

ianagne:
It looks like the spec_helper comes from `crystal init app`.  I created a PR to prepend the ENV to the spec_helper after its created to follow the pattern we're using with gitignore and shards.

ianagne:
I'm super glad this got mentioned.  I'm realizing now that I've wiped my dev database more than once without noticing 😂

paulcsmith:
Closed by #101

PR #101, opened 2017-12-02
Add LUCKY_ENV to spec_helper

ianagne:
Without explicitly setting the environment to "test" in the spec helper,
its pretty easy to accidentally delete your development database when
cleaning up after running specs.

See issue #100

### DIFF
--- a/src/generators/web.cr
+++ b/src/generators/web.cr
@@ -35,6 +35,7 @@ class LuckyCli::Generators::Web
     generate_default_crystal_project
     add_deps_to_shard_file
     remove_generated_src_files
+    remove_generatred_spec_files
     remove_default_readme
     add_default_lucky_structure_to_src
     setup_gitignore
@@ -73,6 +74,10 @@ class LuckyCli::Generators::Web
     FileUtils.rm_r("#{project_dir}/src")
   end
 
+  private def remove_generatred_spec_files
+    FileUtils.rm_r("#{project_dir}/spec")
+  end
+
   private def remove_default_readme
     FileUtils.rm_r("#{project_dir}/README.md")
   end
@@ -105,10 +110,10 @@ class LuckyCli::Generators::Web
     DEPS_LIST
   end
 
-    private def ensure_directory_does_not_exist
-      if Dir.exists?("./#{project_dir}")
-        puts "Folder named #{project_name} already exists, please use a different name"
-        exit
-      end
+  private def ensure_directory_does_not_exist
+    if Dir.exists?("./#{project_dir}")
+      puts "Folder named #{project_name} already exists, please use a different name"
+      exit
     end
+  end
 end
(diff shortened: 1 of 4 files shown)


src/generators/web.cr
```diff
@@ -105,10 +106,16 @@ class LuckyCli::Generators::Web
...
   end
 
-    private def ensure_directory_does_not_exist
-      if Dir.exists?("./#{project_dir}")
-        puts "Folder named #{project_name} already exists, please use a different name"
-        exit
-      end
+  private def add_env_to_spec_helper
+    prepend_text to: "spec/spec_helper.cr", text: <<-LUCKY_ENV
+    ENV["LUCKY_ENV"] = "test"
```
ianagne: I just realized that I deleted the newline characters I had in an earlier version.  Re-adding them now.

paulcsmith:
@ianagne Thanks for tackling this. There are a few other changes that need to be made for specs to work well. Enough that I think it makes sense to remove the default `spec` directory and create our own using teeplate. I'd love it if tackled it and I'm happy to help, but it is a bit bigger, so if you don't have time that is ok.

Here's the gist:

1. Remove the default generated `spec` folder (like we do with removing the default `src` folder.
2. Add a new spec folder by adding it to: https://github.com/luckyframework/lucky_cli/tree/master/src/web_app_skeleton

The new spec *folder* structure should look like this:

```
spec/
  spec_helper.cr
  support/
    .keep
```

The `spec_helper.cr` should look like this:

```crystal
ENV["LUCKY_ENV"] = "test"
require "spec"
require "../src/app"
require "./support/**"
```

> Later on we'll also want to add db truncation to this and some Lucky helpers, but for now this should work.

3. Test that `crystal spec` runs successfully in a newly generated project. It would go here: https://github.com/luckyframework/lucky_cli/blob/223bbb441a33000d94ea7325d809b2f8c7f42581/spec/integration/init_web_spec.cr#L13 with all the other tests

Does that make sense? And if so, is that something you'd be willing to try?

ianagne:
I can definitely tackle this.  I'm doing an OSS Advent of Code, so this is a perfect issue to take on.

paulcsmith:
Cool! I'm so grateful for all the help on this and with the Lucky generators 🎉

src/generators/web.cr
```diff
@@ -73,6 +74,10 @@ class LuckyCli::Generators::Web
     FileUtils.rm_r("#{project_dir}/src")
   end
 
+  private def remove_generatred_spec_files
```
paulcsmith: Small typo in "generated"

paulcsmith: Nevermind, I realized I can do this pretty easily in the GitHub UI after merging.

paulcsmith:
Thanks @ianagne!

src/generators/web.cr
ianagne: Awesome, thanks 👍

PR #107, opened 2017-12-05
Setup Circle CI

paulcsmith:
WIP: I'll have to continue this late since lunch is over

May be helpful https://github.com/heroku/cli-engine-command/blob/fcd620297766a82058b06368eac5d36e2914e961/.circleci/config.yml#L10

https://github.com/search?utf8=✓&q=yarn+path%3A%2F.circleci&type=Code&ref=advsearch&l=&l=

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,20 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: crystallang/crystal
+        environment:
+          DATABASE_URL: postgres://myapp-user@127.0.0.1/myapp-test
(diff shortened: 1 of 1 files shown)


paulcsmith:
Closing. Going to try Travis instead. It seems Circle doesn't support running commands from multiple docker images. So I can either use `yarn` or `crystal` but not both. The way they want me to do it is by making a customer docker image with Crystal and node. We'll see if Travis is simpler

paulcsmith:
Travis worked with 0 setup. Going with that instead

PR #114, opened 2017-12-17
Non-0 exit code when task isn't found

### LINKED ISSUE #108: Exit with non-0 status code if task not found

edwardloveall:
Trying to work on #108 but I've been running into trouble. I'm not at all sure my change did the job and I can't figure out how to test it. I've tried a couple things:

First, I tried adding a test to the runner spec, but it seems that if you call `exit` in the src, it will stop the test immediately so you can't actually check anything.

I ended up with another integration spec that creates a whole lucky app, inserts the new and updated cli shard, and tries to run a non-existing task. This generated an error like this:

```shell
Running my.cool_task
Task my.cool_task not found
F

Failures:

  1) Running a task runs an existing task
     Failure/Error: task("my.cool_task").exit_status.should be_not_found

       test-project/node_modules/buffer/test/constructor.js

     # spec/integration/task_spec.cr:17

Finished in 33.82 seconds
1 examples, 1 failures, 0 errors, 0 pending

Failed examples:

crystal spec spec/integration/task_spec.cr:4 # Running a task runs an existing task
```

I can't figure out what's failing here. Eventually, I want to switch this to a minimal "app" that has lucky_cli as its only dependency and a single empty task to test against. I got that working locally, but the exit code wasn't returning 127. So I'm at a bit of a loss.

FWIW, I chose 127 as the exit code because it [commonly means](https://stackoverflow.com/a/1535733/638966) "command not found".

fixes #108

### DIFF
--- a/spec/integration/task_spec.cr
+++ b/spec/integration/task_spec.cr
@@ -1,0 +1,24 @@
+require "../spec_helper"
+
+describe "Running a task" do
+  it "returns a non-zero exit code when running a missing task" do
+    task("missing.task").exit_status.should_not be_successful
+  end
+end
+
+private def run(process)
+  Process.run(
+    process,
+    shell: true,
+    output: true,
+    error: true
+  )
+end
+
+private def task(task_name)
+  run("crystal spec/tasks.cr --no-debug -- #{task_name}")
+end
+
+private def be_successful
+  eq 0
+end
--- a/spec/lucky_cli/runner_spec.cr
+++ b/spec/lucky_cli/runner_spec.cr
@@ -28,9 +28,15 @@ describe LuckyCli::Runner do
   end
 
   it "does not call the task task is not found" do
-    LuckyCli::Runner
-      .run(args: ["not.real"])
-      .should_not have_called_my_cool_task
+    begin
+      LuckyCli::Runner.exit_with_error_if_not_found = false
+
+      LuckyCli::Runner
+        .run(args: ["not.real"])
+        .should_not have_called_my_cool_task
+    ensure
+      LuckyCli::Runner.exit_with_error_if_not_found = true
+    end
   end
 end
 
--- a/src/lucky_cli/runner.cr
+++ b/src/lucky_cli/runner.cr
@@ -3,6 +3,7 @@ require "./text_helpers"
 
 class LuckyCli::Runner
   @@tasks = [] of LuckyCli::Task
+  class_property exit_with_error_if_not_found = true
 
   extend LuckyCli::TextHelpers
 
@@ -26,7 +27,9 @@ class LuckyCli::Runner
         task.call
       else
         TaskNotFoundErrorMessage.print(task_name)
-        exit(127)
+        if (@@exit_with_error_if_not_found)
+          exit(127)
+        end
       end
     end
   end
--- a/src/lucky_cli/runner.cr
+++ b/src/lucky_cli/runner.cr
@@ -26,6 +26,7 @@ class LuckyCli::Runner
         task.call
       else
         TaskNotFoundErrorMessage.print(task_name)
+        exit(127)
       end
     end
   end
--- a/spec/tasks.cr
+++ b/spec/tasks.cr
@@ -1,0 +1,8 @@
+require "../lucky_cli"
+
+class PlaceholderTask < LuckyCli::Task
+  def call; end
+  def banner; end
+end
+
+LuckyCli::Runner.run


hanneskaeufler:
Regarding stubbing the exit function. This is the closest I got with a tiny bit of fiddling: https://play.crystal-lang.org/#/r/397f

edwardloveall:
Oooo! That is _much_ cleaner. It's almost working but returns error code `32512` which is definitely odd. Any thoughts on that? Technically this completes the issue, but it does feel a bit "gross".

Also, the PlaceholderTask is required because otherwise the task runner doesn't have any tasks registered and therefore doesn't know the type that `tasks_list` should return.

edwardloveall:
Actually, maybe it did exit correctly. From [this](https://stackoverflow.com/questions/5638321/why-child-process-returns-exit-status-32512-in-unix) StackOverflow post:

```
Bits 15-8 = Exit code.
Bit     7 = 1 if a core dump was produced.
Bits  6-0 = Signal number that killed the process.
```
> 32512 = 0x7F00
>So it didn't die from a signal, a core dump wasn't produced, and it exited with code 127 (0x7F).

Still kind of weird? Maybe this is a crystal shortcoming at the moment?

If you want to merge this, let me know so I can A) change the output code, B) clean up the test code a bit and put a nice commit message on it 🙂

spec/integration/task_spec.cr
```diff
@@ -0,0 +1,24 @@
...
+    error: true
+  )
+end
+
+private def task(task_name)
+  run("crystal spec/tasks.cr --no-debug -- #{task_name}")
+end
+
+private def be_not_found
+  eq 127
```
paulcsmith: Since we're not 100% sure about the status code, and I think as long as we test that the exit wasn't `0` what if we change this to `should_not be_successful` and this helper would be `eq 0`

I think that will fix the test, and it will be less coupled to the exact status code, which is good since the most important part is that it is not 0

paulcsmith:
Oh, and could you open an issue about it failing to compile if there are no tasks? I think it should still work if that's the case instead of failing because there are no tasks

spec/integration/task_spec.cr
edwardloveall: Sure thing!

edwardloveall:
Yup, I'll do that after work tonight

paulcsmith:
Thanks!

edwardloveall:
Here's the issue: #116

edwardloveall:
And oh boy, travis failed because it exited with error code 127 🤣 

Hmmm..... This is actually what happens on my machine too. Tests pass, but there's a non-zero exit code. I wonder if a clean up phase is in order where it runs something benign to clean up the error code. I'll poke around. Chime in if you have ideas.

paulcsmith:
Haha well I guess we know for sure it is exiting with 127 :)

The problem is this spec: https://github.com/luckyframework/lucky_cli/blob/33e1dde806b96759d29a8d22b657914aed4f29c1/spec/lucky_cli/runner_spec.cr#L30-L35

If you run just the spec you wrote it passes and exits with 0.

I'm thinking you could add a property: `property skip_exit_if_not_found = false`. Then only `exit(127)` if that property is set to true. So in that spec you can do `runner.skip_exit_if_not_found = true`

What do you think of that solution?

edwardloveall:
Isn't the point of this task to call `exit` _when_ the task isn't found? I guess I could default `skip_exit_if_not_found` to `true` and then set it `false` for that one test.

paulcsmith:
Yeah maybe the naming is confusing. Maybe it should be `property exit_with_error_if_not_found = true` and then the test could set it to false. That might be easier to read/understand

src/lucky_cli/runner.cr
```diff
@@ -3,9 +3,18 @@ require "./text_helpers"
 
 class LuckyCli::Runner
   @@tasks = [] of LuckyCli::Task
+  @@exit_with_error_if_not_found = true
```
paulcsmith: You can use the `class_property` macro to set this up for you :)

edwardloveall:
I'll rebase this real quick

spec/lucky_cli/runner_spec.cr
```diff
@@ -28,9 +28,13 @@ describe LuckyCli::Runner do
   end
 
   it "does not call the task task is not found" do
+    LuckyCli::Runner.exit_with_error_if_not_found = false
+
     LuckyCli::Runner
       .run(args: ["not.real"])
       .should_not have_called_my_cool_task
+
+    LuckyCli::Runner.exit_with_error_if_not_found = true
```
paulcsmith: Can you wrap this in a `begin/ensure` so that it reset to `true` even if it raises or fails for some reason?

src/lucky_cli/runner.cr
edwardloveall: Ah ha! I was wondering if I was just missing that. Thanks!

paulcsmith: You're welcome. And for future reference, you can check out the "Macro summary" in https://crystal-lang.org/api/0.23.1/Object.html for a bunch of other goodies!

src/lucky_cli/runner.cr
```diff
@@ -26,6 +27,9 @@ class LuckyCli::Runner
         task.call
       else
         TaskNotFoundErrorMessage.print(task_name)
+        if (@@exit_with_error_if_not_found)
```
paulcsmith: You could remove the parens and use `exit_with_error_if_not_found` instead of referencing the class variable directly

paulcsmith:
I'll merge this and make the change real quick in the GitHub UI. Thanks @edwardloveall 🎉

PR #126, opened 2017-12-26
Allow usage of DATABASE_URL for any environment

hanneskaeufler:
This makes it easier to setup CI configurations that do not
build the 'production' variant.

E.g. on circle ci I think the way to spin up a setup can be achieved like

```yml
version: 2
jobs:
  build:
    docker:
      - image: whatever/myimage
        environment:
          DATABASE_URL: postgres://blog_test_user@127.0.0.1/blog_test
      - image: postgres:9.6.2-alpine
        environment:
          POSTGRES_DB: blog_test
          POSTGRES_USER: blog_test_user
```

which obviously shouldn't do a `production` build, but the default "localhost" doesn't neccessarily apply. Also good for docker-compose stuff, which I guess circle ci uses anyway?

### DIFF
--- a/src/web_app_skeleton/config/database.cr.ecr
+++ b/src/web_app_skeleton/config/database.cr.ecr
@@ -1,10 +1,10 @@
 database = "<%= crystal_project_name %>_#{Lucky::Env.name}"
 
 LuckyRecord::Repo.configure do
-  if url = ENV["DATABASE_URL"]?
-    settings.url = url
+  if Lucky::Env.production?
+    settings.url = ENV.fetch("DATABASE_URL")
   else
-    settings.url = LuckyRecord::PostgresURL.build(
+    settings.url = ENV["DATABASE_URL"]? || LuckyRecord::PostgresURL.build(
       hostname: "localhost",
       database: database
     )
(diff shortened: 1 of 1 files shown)


src/web_app_skeleton/config/database.cr.ecr
```diff
@@ -1,8 +1,8 @@
 database = "<%= crystal_project_name %>_#{Lucky::Env.name}"
 
 LuckyRecord::Repo.configure do
-  if Lucky::Env.production?
-    settings.url = ENV.fetch("DATABASE_URL")
+  if url = ENV["DATABASE_URL"]?
+    settings.url = url
   else
     settings.url = LuckyRecord::PostgresURL.build(
```
paulcsmith: I like this general idea, but I would like to keep it so that in production you are required to have a DATABASE_URL without it falling back to localhost. So instead, you could do something like this:

```crystal
if Lucky::Env.production?
  settings.url = ENV.fetch("DATABASE_URL")
else
  settings.url = ENV["DATABASE_URL"]? || LuckyRecord::PostgresURL.build(...)
end
```

hanneskaeufler:
Agreed, production should fail in an obvious way

paulcsmith:
Looks great. Thanks, Hannes!

PR #144, opened 2018-01-04
Generate .crystal-version file

### LINKED ISSUE #141: Generate a .crystal-version file
That way when a new version of Crystal comes out, it won't suddenly break build or deploys

edwardloveall:
I would have loved to make this an ecr file that automatially detected the version, maybe from the shard.yml. I wasn't sure how to do that short of parsing the yaml but that seemed like overkill.

fixes #141


paulcsmith:
I think you’re absolutely right. Getting it from the shard file or from the currently installed crystal version would be nice, but this is a good start. 

Do you mind adding an issue for using the user’s installed crystal version?

edwardloveall:
Done: https://github.com/luckyframework/lucky_cli/issues/145

PR #146, opened 2018-01-05
Allow tasks to set a custom name

### LINKED ISSUE #138: Allow custom name
Code is in similar place as #139   

For this issue we should create a macro `name` that accepts a string and generates a `name` method so that users can customize the name if they don’t want the name to be inferred from the class name.

edwardloveall:
If no name is set, it defaults to inferring the name as it does now: by downcasing and replacing the namespace colons with a period.

The two `name` methods inside of `LukcyCLI::Task` strikes me as a little odd, but it does work. I'm not exactly sure how Crystal handles with one to pick. Maybe last implementation wins? Either way I couldn't figure out how to combine them into one method. I needed some way to generate a default name if nothing was specified, but override it with this custom one if it exists, and this is only specified through being inherited.

I also had `abstract def call` in here at one point but it's not needed currently because being inherited always generates the name method, even if you then later pass a custom name in yourself.

fixes #138

### DIFF
--- a/src/lucky_cli/task.cr
+++ b/src/lucky_cli/task.cr
@@ -13,6 +13,18 @@ abstract class LuckyCli::Task
     end
   end
 
+  # Set a custom title for the task
+  #
+  # By default the name is derived from the full module and class name.
+  # However if that name is not desired, a custom one can be set here.
+  # ```
+  # class Dev::Prime < LuckyCli::Task
+  #   name "Development database primer"
+  #   banner "Seed the development database with example data"
+  #
+  #   # other methods, etc.
+  # end
+  # ```
   macro name(name_text)
     def name
       {{name_text}}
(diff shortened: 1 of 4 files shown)


src/lucky_cli/task.cr
```diff
@@ -13,6 +13,12 @@ abstract class LuckyCli::Task
     end
   end
 
+  macro name(name_text)
```
paulcsmith: Actually I just realized this could use some docs. Now that the api is stabilizing, I think it’d be good to have docs in place

edwardloveall: Sweet. I wanted to look into docs for a bit anyway. I’ll get them in tomorrow.

edwardloveall:
FWIW, running `crystal docs` can't compile. I had to run `crystal docs src/lucky_cli.cr`. I think it's compiling all files inside the root of `src` and it goes alphabetically, starting with `dev.cr`. That file creates a `LuckyCli::Task` which hasn't been loaded/required yet, and it crashes.

PR #152, opened 2018-01-12
Fix typo /task.cr/tasks.cr/

unn4m3d:


paulcsmith:
Nice catch. Thanks for the PR!

PR #153, opened 2018-01-12
Require serializers when compiling the app

edwardloveall:
I think serializers were decided on in https://github.com/luckyframework/lucky/issues/280, so this makes sure the folder is included when compiling a project.


paulcsmith:
You're totally right, but I think you may need to update your fork, because this should be on master: https://github.com/luckyframework/lucky_cli/commit/085dca244727a8e789f6e8afe14137081284e20e#diff-3010605d84877d656703ee36800a03c4

edwardloveall:
Oh whoops, yeah. I thought i had it updated. Thanks!

Issue #37, opened 2017-10-21
Add `lucky init {project} --customize`


paulcsmith:
This will walk you through setup and ask if you want certain things

* API only
* Turbolinks
* Heroku
* Brunch

etc. A walkthrough is nice because remembering flags sucks. Flags should still be available though for people that like it, and for use for other CLI tools.

paulcsmith:
Welcome back @mikeeus! I'm glad to hear from you again :D

I'd *love* some help with this. This will be a fairly large feature I think, but it could be done in multiple PRs to keep things simple to review. I'll try to lay out roughly what needs to be done here. I think for starters a `--api` option would be good. Adding the others can come later. As always, feel free to ping me in Gitter if you need any help/advice with stuff

## Some general pointers

* The entry point for initializing a Lucky applications is: https://github.com/luckyframework/lucky_cli/blob/2064e617825b335b058ebaa8fb713035386091a0/src/lucky.cr#L45
* Which calls: https://github.com/luckyframework/lucky_cli/blob/2064e617825b335b058ebaa8fb713035386091a0/src/init_questions.cr#L6
* There is already some support for asking questions in: https://github.com/luckyframework/lucky_cli/blob/2064e617825b335b058ebaa8fb713035386091a0/src/init_questions.cr#L10-L18 but it is not documented and should be ignored for now. I'll expand it and document it later. This will be what the future `--customize` option uses.
* Generating a Lucky project is done by: https://github.com/luckyframework/lucky_cli/blob/master/src/generators/web.cr
* The templates for the `src` folder are here: https://github.com/luckyframework/lucky_cli/tree/master/src/web_app_skeleton

## With that info, here's roughly what needs to change

* https://github.com/luckyframework/lucky_cli/blob/master/src/generators/web.cr#L6-L10 should set an `@api_only` instance var to true if the user passes `--api`. You can check that directly in `ARGV` or use Crystal's [OptionParser](https://crystal-lang.org/api/0.24.2/OptionParser.html). I'm not sure which will be easiest. It's up to you :)
* Pass the `@api_only` option to the `SrcTemplate` class: `https://github.com/luckyframework/lucky_cli/blob/2064e617825b335b058ebaa8fb713035386091a0/src/generators/web.cr#L72-L74`
[...]

PR #175, opened 2018-04-01
Add `lucky init {project} --api`

mikeeus:
As discussed in #37.

I've done the following:
- Catch options init options in `src/generators/web.cr`, set `@api_only` and pass it to `SrcTemplate`
- Moved browser specific folders and files to `browser_app_skeleton`
- Use `@api_only` to inherit from the correct action in `actions/home/index.cr`
- Use `@api_only` to conditionally require folders in `app.cr`

I've started writing a spec for the `--api` only option, but I'm having trouble passing flags through to the compiled app. The `-D` flag should do this right?
```crystal
should_run_successfully "crystal src/lucky.cr init test-project -D --api"
```


mikeeus:
Alright, I got the test running by building the binary first then initializing the api only app. I also moved back the `bs-config.js` and `package.json` so `browser-sync` could watch for changes. 

One thing to note is that `lucky dev` still tries to run `bin/compile_assets` which I moved to the `browser_app_skeleton`.

It's looking pretty good to me so far. What do you think?

paulcsmith:
I just took a quick look and things are looking good. I'll have to review more in-depth later, but here are some quick ideas:

* I believe you can pass `-- --api` to pass args in the spec
* I think you can remove yarn from bin/setup and remove bin/compile_assets if you also remove the `assets` line from the generated `Procfile.dev`. You may also need to make a change to the `Dev` task in this repo that checks if browsersync is installed or something. I'm not sure how to do that well quite yet so if this is too much I'm ok leaving it as-is for now.

src/lucky_cli/browser_src_template.cr
```diff
@@ -0,0 +1,8 @@
+require "random/secure"
+
+class BrowserSrcTemplate < Teeplate::FileTree
+  directory "#{__DIR__}/../browser_app_skeleton"
+
+  def initialize
+  end
```
paulcsmith: I believe you can remove this. There is a default `initialize` added by Crystal

src/web_app_skeleton/src/actions/errors/show.cr.ecr
```diff
@@ -0,0 +1,52 @@
...
+    if json?
+      json Errors::ShowSerializer.new(message), status: 500
+    else
+      render_error_response status: 500
+    end
+    <%- end -%>
+  end
+
+  private def render_error_response(status : Int32, title : String = "We're sorry. Something went wrong.")
+    <%-if @api_only -%>
```
paulcsmith: I think if it is api only you can remove this method completely since it is no longer used

mikeeus:
K I removed the `initialize` and `render_error_page` methods, passed `-- --api` directly and removed the `assets` from `Procfile.dev`.

I initially removed yarn and browser-sync but I had to re-add them because `lucky watch` uses browser-sync to watch the app and reload the browser.

https://github.com/luckyframework/lucky/blob/8f3081e29c346f04bf2c7e9c36a419ffab0f49e6/tasks/watch.cr#L41-L45

I think I'll also need to make changes in the Lucky repo, for example when generating actions the `ActionTemplate` renders text, but it should render json for a json api.

https://github.com/luckyframework/lucky/blob/8f3081e29c346f04bf2c7e9c36a419ffab0f49e6/tasks/gen/templates/action/%7B%7Baction%7D%7D.cr.ecr#L1-L5

How can we pass `@api_only` to the lucky tasks? Or can we set it in the server settings in `config/server.cr.ecr` and check it from the task?

paulcsmith:
Nice! I'll take a look again soon. Regarding your questions:

### Action generators

* Rename `lucky gen.action` to `lucky action.browser`
* Add `lucky action.api`

These would likely render the same templates, but expose a `@inherit_from` variable that sets either `BrowserAction` or `ApiAction`

### Watch task

I think that it should parse the options and one of those options should be `--reload-browser`. The Watch task would only do the BrowserSync stuff if the `--reload-browser` flag was set. The `lucky init` generator would then only add that flag if `@api_only` is not set

What do you think? Does that make sense?

mikeeus:
I feel like the generators are intuitive with the form `lucky gen.{template}` and I'm not sure how `lucky action.api` looks. What do you think about something like `lucky gen.action_api` and `lucky gen.action_browser`?

As for the watch task that makes perfect sense, I'll create an issue for it in the lucky repo.

paulcsmith:
Yeah you’re right. Let’s keep the “gen”. I think gen.action.api and gen.action.browser make the most sense though. So it’s consistent with dot usage. What do you think?

mikeeus:
Yeah I think that sounds good. I'll work on it tonight and tomorrow morning

mikeeus:
Hi Paul, what else do I need to do to complete pull request?

paulcsmith:
@mikeeus Thanks for the ping! I'm in all day meetings for a design sprint this week so I've barely checked my email or PRs. I'll try to get to this today or Friday

mikeeus:
No problem, I was really busy last week to so I understand.

PR #179, opened 2018-04-07
Configure process runner to use STDIN

### LINKED ISSUE #178: `lucky mytask` break `gets` and `read_line`
Say you have a task lucky app with a task in `tasks/deploy_db.cr` that looks like:

```crystal
require "lucky_cli"

class DeployDb < LuckyCli::Task
  banner "Export database and prepare to mirror it to heroku"

  def call
    puts "Paste the dropbox link now:"
    foobar = gets
    puts foobar
  end
end
```

Running `lucky deploy_db` will return immediately, without waiting for input like it should with `gets`. Using `read_line` instead results in the following error.

bildschirmfoto 2018-04-07 um 16 43 04

Interestingly, running `crystal run ./tasks.cr -- deploy_db` works as intended. So it seems _something_ is emitting a newline (?) to stdin when using `lucky deploy_db` which causes gets to return?

```
16:57:01-~ (master)$ lucky --version
0.9.2
16:57:04-~ (master)$ crystal --version
Crystal 0.24.2 (2018-03-10)

LLVM: 5.0.1
Default target: x86_64-apple-macosx
```

hanneskaeufler:
- [x] Add test

When wanting to use `gets` or `read_line`, the task would immediately
return without accepting input. That is because `Process.run` is
configured to use `Redirect::Close`. We can configure it to accept STDIN
instead.
[Docs for Process.run](https://crystal-lang.org/api/0.24.1/Process.html#run%28command%3AString%2Cargs%3Dnil%2Cenv%3AEnv%3Dnil%2Cclear_env%3ABool%3Dfalse%2Cshell%3ABool%3Dfalse%2Cinput%3AStdio%3DRedirect%3A%3AClose%2Coutput%3AStdio%3DRedirect%3A%3AClose%2Cerror%3AStdio%3DRedirect%3A%3AClose%2Cchdir%3AString%3F%3Dnil%29%3AProcess%3A%3AStatus-class-method)

Closes #178

### DIFF
--- a/spec/integration/task_spec.cr
+++ b/spec/integration/task_spec.cr
@@ -12,6 +12,14 @@ describe "Running a task" do
 
     io.to_s.should eq "Ran precompiled task\n"
   end
+
+  it "allows tasks to accept input from STDIN" do
+    io = IO::Memory.new
+
+    run("echo 'hello world' | LUCKY_TASKS_FILE=./spec/tasks.cr crystal src/lucky.cr -- task_with_input", output: io)
+
+    io.to_s.should eq "input: hello world\n"
+  end
 end
 
 private def run(process, output = STDOUT)
--- a/spec/tasks.cr
+++ b/spec/tasks.cr
@@ -6,4 +6,13 @@ class PlaceholderTask < LuckyCli::Task
   def banner; end
 end
 
+class TaskWithInput < LuckyCli::Task
+  banner "this should be first"
+
+  def call
+    input = gets
+    puts "input: #{input}"
+  end
+end
+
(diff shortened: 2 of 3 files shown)


hanneskaeufler:
Those tests are failing in the same way locally as well. Might check that out later.

paulcsmith:
Yeah, I just noticed that as well. This is a solid start though :D

paulcsmith:
Actually, the same failure is on master. I'll merge this in and figure out the test later!

hanneskaeufler:
Rebased and passing tests now.

PR #218, opened 2018-07-03
Add spec for deploying to Heroku

paulcsmith:
To-do:

- [x] Disable deploy specs by default. Require a flag
- [x] Use a unique id at end of app name so they don't collide
- [x] Loop through apps and delete *all* apps that match the first part of the app name (so you don't have tons of apps for integration specs
- [x] Document how to run these locally
- [x] Remove redundant export path or mv to /usr/local/bin. Choose one


paulcsmith:
Still needs a lot of cleanup and some some work to pass in Travis CI

PR #223, opened 2018-07-12
Generate JWT Auth for Api Applications

mikeeus:
I put all the JWT authentication related methods in an `Auth::ActionHelpers` module and used that in `Auth::RequireSignIn` in the place of `Authentic::ActionHelpers`. I think since most of the session/token/jwt specific stuff is in this module, we can switch it out pretty easily to change authentication methods.

I also added `AppVisitor` and generated specs for `SignIn::Create`, `SignUp::Create`, `Home` and `Me` actions. I made `SignIn` and `SignUp` singular since there is only a create action.


mikeeus:
One thing to note is that I stubbed `Api::PasswordResets::New` so that it just returns text. This is because the email template needs a url to send for the password reset, but that url should point to the client, not the api. What are your thoughts on that?

Also I think generating the `AppVisitor` and adding api specs serves as a good starting point for people to test their apps. It also acts as documentation for how to make authenticated requests, use Boxes, etc.

src/api_authentication_app_skeleton/spec/api_spec.cr
```diff
@@ -0,0 +1,58 @@
+require "./spec_helper"
```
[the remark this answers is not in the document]
paulcsmith: Oh I just realized this spec kind of handles multiple things. I think this should be split up a resource:

```
spec/requests/api/
  - sign_ins_spec.cr
  - sign_ups_spec.cr
  - me_spec.cr
  - password_resets_spec.cr
```

src/api_authentication_app_skeleton/spec/support/app_visitor.cr
```diff
@@ -0,0 +1,131 @@
...
+
+class AppVisitor
+  include Auth::ActionHelpers
+
+  CSRF_TOKEN = "statictesttoken"
+
+  getter! response
+  @response : HTTP::Client::Response? = nil
+
+  def visit(path : String, as_user : Authentic::PasswordAuthenticatable? = nil, headers : HTTP::Headers? = nil)
```
paulcsmith: Let's rename this to `get` so it's more consistent with the other HTTP methods and could you rename `as_user` to `as` to match LuckyFlow?

src/api_authentication_app_skeleton/spec/support/app_visitor.cr
```diff
@@ -0,0 +1,131 @@
...
+  def response_body
+    body = @response.not_nil!.body
+    if body[0, 5] == "ERROR"
+      raise Exception.new(body)
+    elsif body == ""
+      JSON.parse("{}")
+    else
+      JSON.parse(body)
+    end
+  end
```
paulcsmith: I think `response_body` should return the raw_response and there should be another method called `response_json` that parses it. That way this method can work for HTML/XML or whatever people want

src/api_authentication_app_skeleton/spec/support/app_visitor.cr
```diff
@@ -0,0 +1,131 @@
...
+      headers.add "Authorization", generate_token(as_user)
+    end
+
+    request = HTTP::Request.new("DELETE", path, headers)
+    @response = process_request(request)
+  end
+
+  def response_body
+    body = @response.not_nil!.body
+    if body[0, 5] == "ERROR"
```
paulcsmith: Could you check `response.success?` instead of matching the string?

src/api_authentication_app_skeleton/spec/support/app_visitor.cr
```diff
@@ -0,0 +1,131 @@
...
+      raise Exception.new(body)
```
paulcsmith: I think this should raise something like `raise "Return failed response with status code: #{status_code}" so you know why it raised. Thoughts?

src/api_authentication_app_skeleton/src/actions/api/me/show.cr
```diff
@@ -0,0 +1,5 @@
+class Api::Me::Show < ApiAction
+  get "/api/me" do
+    json Users::ShowSerializer.new(current_user)
```
paulcsmith: 👍 I like that this show people how to create and use serializers!

src/api_authentication_app_skeleton/spec/support/app_visitor.cr
```diff
@@ -0,0 +1,131 @@
...
+      Lucky::HttpMethodOverrideHandler.new,
+      Lucky::LogHandler.new,
+      Lucky::SessionHandler.new,
+      Lucky::Flash::Handler.new,
+      Lucky::ErrorHandler.new(action: Errors::Show),
+      Lucky::RouteHandler.new,
+      Lucky::StaticFileHandler.new("./public", false),
+      Lucky::RouteNotFoundHandler.new,
+    ])
+  end
```
paulcsmith: I've been thinking about this and I think it is a bit brittle. It relies on copying the middleware and doesn't fully test how a web request comes through the server. I think instead it would be better to use [`HTTP::Client`](https://crystal-lang.org/api/0.25.1/HTTP/Client.html) to make a real request against the server. 

You can use Lucky::RouteHelper.settings.base_uri to get the base_uri for the request.

For this to work you'll also need to move: https://github.com/luckyframework/lucky_cli/blob/master/src/browser_app_skeleton/spec/setup/start_app_server.cr to `web_app_base_skeleton` so it is available to browser and api apps. It will need to conditionally shutdown LuckyFlow for just browser apps. Does this make sense?

It's kind of complicated so LMK if I should explain some more

src/api_authentication_app_skeleton/src/actions/api/password_reset_requests/create.cr
```diff
@@ -0,0 +1,14 @@
+class Api::PasswordResetRequests::Create < ApiAction
+  include Auth::RequireNotSignedIn
+
+  route do
+    PasswordResetRequestForm.new(params).submit do |form, user|
+      if user
+        PasswordResetRequestEmail.new(user).deliver
+        json({message: "You should receive an email on how to reset your password shortly"}, Status::OK)
```
paulcsmith: I believe you can do `:ok` instead of `Status::OK`. I'm not sure though. Also if we use the enums here I think they should be used everywhere so it isn't mixing and matching enums with raw int status codes

src/api_authentication_app_skeleton/src/actions/api/password_resets/new.cr
```diff
@@ -0,0 +1,9 @@
+class Api::PasswordResets::New < ApiAction
+  include Auth::PasswordResets::Base
+
+  param token : String
+
+  get "/api/password_resets/:user_id" do
+    text "Please configure password reset page on client."
+  end
+end
```
paulcsmith: I think this can be removed and instead we can add something to the authentication guide: https://luckyframework.org/guides/authentication/

src/api_authentication_app_skeleton/src/actions/api/sign_in/create.cr
```diff
@@ -0,0 +1,14 @@
+class Api::SignIn::Create < ApiAction
```
paulcsmith: To make it consistent can this be pluralized? Even though it's only used once that is the typical style.

src/api_authentication_app_skeleton/src/actions/api/sign_in/create.cr
```diff
@@ -0,0 +1,14 @@
...
+  include Auth::RequireNotSignedIn
```
paulcsmith: Can you namespace all the API specific auth stuff in `Auth::Api`, example: `Auth::Api::RequireNotSignedIn`. That way they don't collide with the ones for browsers

src/api_authentication_app_skeleton/src/actions/mixins/auth/action_helpers.cr
```diff
@@ -0,0 +1,30 @@
+require "jwt"
+
+module Auth::ActionHelpers
```
paulcsmith: This could probably be namespaced too since it is API specific. Otherwise it will clobber the browser-based auth helpers

src/api_authentication_app_skeleton/src/actions/mixins/auth/action_helpers.cr
```diff
@@ -0,0 +1,30 @@
...
+  rescue KeyError # Missing hash key: "Authorization" (KeyError)
+    nil
+  rescue JWT::ExpiredSignatureError
+    nil
+  end
+
+  def generate_token(user : Authentic::PasswordAuthenticatable)
+    payload = {"sub" => user.id, "exp" => 14.days.from_now.epoch}
+    JWT.encode(payload, Lucky::Server.settings.secret_key_base, "HS256")
+  end
```
paulcsmith: What do you think about moving this to the `User` model so the module doesn't need to be included in so many places? It seems this is something that is useful outside HTTP actions o I think the model or a separate `TokenGenerator` object in `app/models` would work. Since it's so simple I think a method in User would be best though

src/generators/web.cr
```diff
@@ -174,6 +177,14 @@ class LuckyCli::Generators::Web
           version: ~> 0.2
       DEPS_LIST
     end
+
+    if api_only? && generate_auth?
+      append_text to: "shard.yml", text: <<-DEPS_LIST
+
```
paulcsmith: If you rebase there should be a fix that adds a newline so I don't this this newline is necessary anymore

paulcsmith:
This is really incredible work. This would have taken me so much time. Thank you for doing it. Can't wait to get this in 👍

src/api_authentication_app_skeleton/spec/api_spec.cr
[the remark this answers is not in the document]
mikeeus: I agree putting it under `spec/requests` makes sense.

src/api_authentication_app_skeleton/spec/api_spec.cr
```diff
@@ -0,0 +1,58 @@
...
+
+include Auth::ActionHelpers
+
+describe App do
+  visitor = AppVisitor.new
```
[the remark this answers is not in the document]
mikeeus: Yeah I like it, it reads nice

src/api_authentication_app_skeleton/spec/support/app_visitor.cr
mikeeus: Sure, I originally had it as `as` but I was worried that it might be a restricted keyword. If there is no issues then it definitely looks better

src/api_authentication_app_skeleton/src/actions/api/password_resets/new.cr
mikeeus: Okay but if I remove this action it raises an error since the email templates at `/src/base_authentication_app_skeleton/src/emails/templates/password_reset_request_email/text.ecr.ecr` reference the action:

```
Please reset your password:

<%="<"%>%= PasswordResets::New.url(@user.id, @token) %>
```

src/api_authentication_app_skeleton/spec/support/app_visitor.cr
mikeeus: Yes I see what you mean. I'll try to set it up

mikeeus:
Alright, I think I made all the necessary changes. Can you review it again?

src/api_authentication_app_skeleton/spec/requests/api/home_spec.cr
```diff
@@ -0,0 +1,11 @@
...
+
+describe "Home" do
+  it "says hello" do
+    request = AppRequest.new
+    
+    request.get("/")
+
+    request.response_json.should eq ({"hello" => "Hello World from Home::Index"})
+  end
+end
```
paulcsmith: This is missing a newline. You may want to run `crystal tool format src spec` to format everything

src/api_authentication_app_skeleton/spec/requests/api/me_spec.cr
```diff
@@ -0,0 +1,19 @@
+require "../../spec_helper"
+
+describe "Me" do
+  it "returns current user" do
+    request = AppRequest.new
+
+    user = UserBox.create
```
paulcsmith: Can you put this on the line above in the "setup" phase

src/api_authentication_app_skeleton/spec/requests/api/sign_ups_spec.cr
```diff
@@ -0,0 +1,20 @@
...
+      "sign_up:email"                 => "test@email.com",
+      "sign_up:password"              => "password",
+      "sign_up:password_confirmation" => "password",
+    }))
+
+    request.response.status_code.should eq 200
+
+    new_user = UserQuery.new.email("test@email.com").first
+    new_user.should_not be_nil
+
```
paulcsmith: I think these lines 14 and 17 can be removed. You may also want move it above line 13 so the expectations are all in a row

src/api_authentication_app_skeleton/spec/support/app_request.cr
```diff
@@ -0,0 +1,83 @@
...
+      body_strings << "#{URI.escape(key)}=#{URI.escape(value)}"
+    end
+    body_strings.join("&")
+  end
+
+  private def hash_to_headers(hash : Hash(String, String))
+    headers = HTTP::Headers.new
+    hash.keys.map { |key| headers.add key, hash[key] }
+    headers
+  end
```
paulcsmith: I think it would be awesome to extract this to `lucky` itself as `Lucky::Request`. Then in lucky_cli create a `AppRequest < Lucky::Request`. That way if people want they can add their own methods to `AppRequest` and we can make bug fixes in Lucky without people needing to update the generated file. Thoughts?

src/api_authentication_app_skeleton/src/actions/api/password_reset_requests/create.cr
paulcsmith: What do you think about this? I think all status codes should use symbols if possible

src/api_authentication_app_skeleton/src/actions/api/password_resets/new.cr
paulcsmith: I'm beginning to think that we should remove PasswordResets for now, or start them in a separate PR after this one is removed. It's not quite there IMO and I'd rather not have it block the rest of this PR

It would be a bit of a pain for people to implement themselves, but I think it'd also be a pain to figure out how to integrate with it as it is now. So I think we should hold off on it

What do you think?

src/api_authentication_app_skeleton/src/actions/mixins/auth/api/action_helpers.cr
```diff
@@ -0,0 +1,23 @@
...
+    context.response.headers.add "Authorization", user.generate_token
+  end
+
+  def token
+    context.request.headers["Authorization"]
+  end
+
+  def current_user?
+    user_from_token(token.not_nil!)
+  rescue KeyError | JWT::DecodeError | JWT::ExpiredSignatureError
```
paulcsmith: Rescuing KeyError can hide some nasty bugs. It's a fairly generic error so it could be raised if the Authorization header is missing, but it might also raise if any other key is missing. Instead I think something like this would be safer:

```crystal
def current_user?
  token.try do |value|
    user_from_token(value)
  end
rescue JWT::DecodeError | JWT::ExpiredSignatureError
  nil
end
```

mikeeus:
Nice, this is coming along! Thanks a lot for taking the time to review this, I know its a lot to go through. It's late for me but I'll try to finish up the remaining issues tomorrow.

src/api_authentication_app_skeleton/src/actions/mixins/auth/api/action_helpers.cr
mikeeus: Yeah that makes sense, I'll take it out

mikeeus: Okay I though I could use `.try` like this:
```
def current_user?
  token.try do |value|
    user_from_token(value)
  end
rescue JWT::DecodeError | JWT::ExpiredSignatureError
  nil
end
```
But it still throws `"ERROR: Missing hash key: HTTP::Headers::Key(@name=\"Authorization\") (KeyError)\n"` since token is trying to access the `Authorization` header:
```crystal
def token
  context.request.headers["Authorization"]
end
```

That's why I caught the `KeyError` before. What if I rescued the `KeyError` in the token method? like:

```crystal
def token
  context.request.headers["Authorization"]
  rescue KeyError
    nil
end
```
Since I'm only trying to access the `Authorization` key it shouldn't hide any other possible errors right?

paulcsmith: Ah I see. I think what you want is `...headers["Authorization"]?` (note the trailing `?`). That will tell the compiler it can return `nil` without raising. LMK if that does the trick

paulcsmith: Oh it looks like you did this. I didn't refresh so this still looked open :)

mikeeus:
No problem, I'll work on other issues and make new PRs in the meantime. Good luck on your project

jackturnbull:
Nothing to add in terms of the PR but this is something I'll find really useful. Thank you putting the time in.

paulcsmith:
https://github.com/luckyframework/lucky_cli/pull/234 fixes the Heroku specs. Could you rebase and force push so it can re-run the specs with the latest changes on master?

src/api_authentication_app_skeleton/spec/support/app_request.cr
paulcsmith: I think it depends a bit on how much time you have. If you are open to it, I think adding this to Lucky before merging this PR would be great. And then this PR could use it.

If you don't have much time then yeah we can make an issue for it and extract it later

src/base_authentication_app_skeleton/src/emails/templates/password_reset_request_email/html.ecr.ecr
```diff
@@ -1,3 +1,8 @@
 <h1>Please reset your password</h1>
 
+<%- if api_only? -%>
+<a href="<%="<"%>%= Api::PasswordResets::New.url(@user.id, @token) %>">Reset password</a>
```
paulcsmith: I think the conditional should be removed and this and the other password reset stuff should be moved to the browser specific auth folder.

We can tackle password reset later when we have really nailed down how it should work. Does that sound good to you?

src/web_app_skeleton/src/actions/api_action.cr.ecr
```diff
@@ -0,0 +1,18 @@
+abstract class ApiAction < Lucky::Action
+  <%- if generate_auth? -%>
+  # By default all actions require sign in, unless you use the
+  # `Auth::SkipRequireSignIn` module in `src/mixins/auth/skip_require_sign_in.cr`
+  <%- if api_only? -%>
+  include Auth::Api::RequireSignIn
+  <%- else -%>
+  include Auth::RequireSignIn
+  <%- end -%>
```
paulcsmith: Since this is an ApiAction I don't think it needs the the `api_only?` conditional part. I think it should always use `Auth::Api::RequireSignIn`

src/web_app_skeleton/src/actions/home/index.cr.ecr
```diff
@@ -1,5 +1,9 @@
 <%- if api_only? -%>
 class Home::Index < ApiAction
+  <%- if generate_auth? -%>
+  include Auth::SkipRequireSignIn
```
paulcsmith: Should this be using the `Api` version?

mikeeus:
I've been testing the PR by generating api applications and running the specs to ensure they work. But I'm uncomfortable about the travis CI tests not passing. Whats the issue with that? Can I help?

I'll make an issue for Lucky::Request and adding that first. I can probably get that done tomorrow.

paulcsmith:
@mikeeus Good point about CI. Can you open a PR with a super minor change like adding `puts "Hi"` somewhere in a spec? It *should* pass just fine, but it would be nice to know for sure. That'll help us track down if it is a configuration issue or something in this PR that is breaking things

paulcsmith:
@mikeeus If you rebase and force push Travis should work now. Here's the PR that fixed it, along with an explanation https://github.com/luckyframework/lucky_cli/pull/238

LMK if it works!

mikeeus:
Hey @paulcsmith I just rebased and pushed it up

src/base_authentication_app_skeleton/src/emails/templates/password_reset_request_email/html.ecr.ecr
paulcsmith: Bump

src/web_app_skeleton/src/actions/api_action.cr
```diff
@@ -1,3 +0,0 @@
-abstract class ApiAction < Lucky::Action
-  # Add pipes and methods that are for all API requests
-end
```
paulcsmith: I think this can be left in. Full web apps should also have an ApiAction

src/base_authentication_app_skeleton/src/emails/templates/password_reset_request_email/html.ecr.ecr
mikeeus: Ah yes, I had totally forgotten about this. Thanks for bringing it to my attention.

src/web_app_skeleton/src/actions/api_action.cr
mikeeus: I think this was renamed to the `api_action.cr.ecr` file below

src/web_app_skeleton/src/actions/api_action.cr.ecr
mikeeus: So that means that browser apps would need to create a separate `RequireSignIn` mixin for their api actions, instead of using the one that is generated for them right?

paulcsmith: Hmm good point. There's what I think this should be:

```crystal
<%- if generate_auth? -%>
include Auth::Api::RequireSignIn
<%- end %> 
```

And the `else` should be omitted. The ApiAction is always created, so the only conditional would be whether it should generate auth or not. So that means the `Auth::ApiRequireSignIn` module should be created for full or api_only apps, but only if generated_auth? is true

src/web_app_skeleton/src/actions/home/index.cr.ecr
paulcsmith: I guess it doesn't matter actually

mikeeus: You're right, I'm not sure since this action is at the `/` route and not at `/api/`. If you want I can remove the api version and just leave the lucky welcome page.

src/web_app_skeleton/src/actions/api_action.cr.ecr
mikeeus: Yes that makes sense

mikeeus:
Okay so it turns out I do need to add a case for `api_only?` in the `api_action` because there is no `Auth::Api::RequireSignIn` module in the browser app which needs to use `Auth::RequireSignIn` instead.

paulcsmith:
I think Auth::Api::RequireSignIn should also be available for browser apps since browser apps also have an api action. So then you wouldn’t need the api only check

The reason is that Auth::RequireSignIn  does a redirect, which isn’t what you’d want in an API response. So for full Lucky apps you’d want Auth::ApiRequireSignIn

This is also making me thing that the generated specs should also be in the full browser app too because you’d want those either way. Is this making sense? This is a tricky new feature so maybe I’m misunderstanding something too

src/generators/web.cr
```diff
@@ -45,10 +45,9 @@ class LuckyCli::Generators::Web
       add_browser_authentication_to_src
     end
 
-    # TODO: Add API auth with JWT
-    # if api_only? && generate_auth?
-    #   add_api_authentication_to_src
-    # end
+    if api_only? && generate_auth?
+      add_api_authentication_to_src
+    end
```
paulcsmith: So I think my original todo snippet was wrong. This could be generated even without api_only? Because you’d want the API auth  in browser apps too.

We can make it optional in the “init” task later but I think that makes the most sense. What do you think?

mikeeus: Okay that means that there would be normal (session) authentication and jwt authentication in the browser app right? I guess it does make sense since the api will need authentication too

mikeeus: I see what you mean now. If we have the api being generated for the browser apps then it should get authentication generated too. I made the change and pushed up  to see if the tests pass

src/api_authentication_app_skeleton/spec/support/app_request.cr
HarrisonB: Need a hand? Happy to help with that migration to `lucky`!

mikeeus:
@paulcsmith I think this PR is ready for review. I refactored `AppRequest` into `Lucky::Request`, do we need to [merge that PR](https://github.com/luckyframework/lucky/pull/600) before we finish this one?

jwoertink:
Hey all, I'm curious the status on this. Looks like we're going to be doing some JWT stuff soon, so would be nice to have this stuff all in there.

paulcsmith:
@jwoertink I'm going to review https://github.com/luckyframework/lucky/pull/689 and then I'll fork this and finish it up. I'm not exactly sure when. When do you think you'd need this?

Also I'm wondering if we need to use a JWT or if we can use a regular token in the Authorization header. I'm not sure JWT should be the default. What do you think?

jwoertink:
I've never used JWT before. I don't know anything about it other than it seems to be what a lot of people are using. We're currently working on 2 lucky apps where one of them is just a json API for the other one. Maybe this could just be extracted out in to a separate shard? There probably still needs to be a few updates to make for easier integration like this AppRequest stuff?

paulcsmith:
Yeah I'll need to update it to use `AppRequest`. I think it can be built-in to the CLI and I think using JWT would be pretty easy to change but I'm not sure. I'll keep you posted by commenting here as things move forward. Would definitely love feedback on it as I move forward!

paulcsmith:
Closing in favor of #395. Thanks again @mikeeus. This was a big help in getting this feature in. I hope life is treating you well!

PR #236, opened 2018-07-31
Check if it puts hi

mikeeus:


mikeeus:
The question is: will it puts hi?

paulcsmith:
Thanks for posting this. I'll try to fix it tonight so external contributions don't always fail

PR #250, opened 2018-08-21
Sanitize name in lucky init

### LINKED ISSUE #248: lucky init should strip invalid characters from project name
I was doing some cli hijinks trying to bootstrap a project with lucky into an existing folder and ended up running `lucky init project_name/` which had a bunch of weird side effects, for example:

```yml
# shard.yml
name: project_name/
version: 0.1.0

targets:
  project_name/:
    main: src/project_name/.cr

crystal: 0.26.0
```

robacarp:
fixes #248 

Quick and simple fix for #248. There could probably be a _little_ more logic here, but I'm not sure what that'd include. Since I don't see Lucky building any class names with the project name, I think this is fine.

It may be worth collapsing multiple `_`s.

The distinction between `@project_name` and `@project_dir` may be unnecessary now, or I may be incorrect in making them the same.

This sort of thing is difficult write tests for, but here's what I did in the terminal. Minimally edited to protect the name of the innocent computer :P

```text
> rm -r yolo; lucky_cli/bin/lucky init Yolo/; ag -i yolo yolo
Done generating your Lucky project

  ▸ cd into yolo
  ▸ check database settings in config/database.cr
  ▸ run bin/setup
  ▸ run lucky dev to start the server
yolo/config/database.cr
1:database = "yolo_#{Lucky::Env.name}"

yolo/config/session.cr
4:  settings.key = "yolo"

yolo/shard.yml
1:name: yolo
8:  yolo:
9:    main: src/yolo.cr

yolo/README.md
1:# yolo

##########################################################################
> rm -r yolo; lucky_cli/bin/lucky init Yolo/////; ag -i yolo yolo/
Done generating your Lucky project

  ▸ cd into yolo
  ▸ check database settings in config/database.cr
  ▸ run bin/setup
  ▸ run lucky dev to start the server
yolo/config/database.cr
1:database = "yolo_#{Lucky::Env.name}"

yolo/config/session.cr
4:  settings.key = "yolo"

yolo/shard.yml
1:name: yolo
8:  yolo:
9:    main: src/yolo.cr

yolo/README.md
1:# yolo

##########################################################################
> rm -r yolo; lucky_cli/bin/lucky init Yolo/////; ag -i yolo 
Done generating your Lucky project

  ▸ cd into yolo
  ▸ check database settings in config/database.cr
  ▸ run bin/setup
  ▸ run lucky dev to start the server
yolo/config/database.cr
1:database = "yolo_#{Lucky::Env.name}"

yolo/config/session.cr
4:  settings.key = "yolo"

yolo/shard.yml
1:name: yolo
8:  yolo:
9:    main: src/yolo.cr

yolo/README.md
1:# yolo

##########################################################################
> rm -r yolo; lucky_cli/bin/lucky init 'Yol\(\*&\(\*&o/////'; ag -i yolo 
Done generating your Lucky project

  ▸ cd into yol__________o
  ▸ check database settings in config/database.cr
  ▸ run bin/setup
  ▸ run lucky dev to start the server

> ag -i yol
yol__________o/config/database.cr
1:database = "yol__________o_#{Lucky::Env.name}"

yol__________o/config/session.cr
4:  settings.key = "yol__________o"

yol__________o/shard.yml
1:name: yol__________o
8:  yol__________o:
9:    main: src/yol__________o.cr

yol__________o/README.md
1:# yol__________o

##########################################################################
> rm -r yolo; lucky_cli/bin/lucky init 'Yol23/'; ag -i yolo 
rm: yolo: No such file or directory
Done generating your Lucky project

  ▸ cd into yol23
  ▸ check database settings in config/database.cr
  ▸ run bin/setup
  ▸ run lucky dev to start the server

> ag yol23
yol23/config/database.cr
1:database = "yol23_#{Lucky::Env.name}"

yol23/config/session.cr
4:  settings.key = "yol23"

yol23/shard.yml
1:name: yol23
8:  yol23:
9:    main: src/yol23.cr

yol23/README.md
1:# yol23
```

### DIFF
--- a/spec/integration/init_web_spec.cr
+++ b/spec/integration/init_web_spec.cr
@@ -68,6 +68,30 @@ describe "Initializing a new web project" do
     message = "Folder named test-project already exists, please use a different name"
     output.to_s.strip.should contain(message)
   end
+
+  it "does not create project if project name is not a valid project name" do
+    output = IO::Memory.new
+    Process.run(
+      "crystal src/lucky.cr init 'test project'",
+      output: output,
+      shell: true
+    )
+    message = "Project name should only contain letters, numbers, underscores, and dashes."
+    output.to_s.strip.should contain(message)
+  end
+
+  it "translates dashes to underscores in the project name" do
+    output = IO::Memory.new
+    Process.run(
+      "crystal src/lucky.cr init 'test-project'",
+      output: output,
+      shell: true
+    )
+
+    shard_yml_name = File.read_lines("test-project/shard.yml").select { |line| line =~ /^name:/ }.first
+    shard_yml_name.should eq("name: test_project")
+    File.exists?("test-project/src/test_project.cr").should be_true
+  end
 end
 
 private def compile_and_run_specs_on_test_project
--- a/spec/lucky_cli/generators/validators_spec.cr
+++ b/spec/lucky_cli/generators/validators_spec.cr
@@ -1,0 +1,37 @@
+require "../../spec_helper"
+
+private def validate(string)
+  LuckyCli::Generators::Validators::ProjectName.valid? string
+end
+
+private def sanitize(string)
+  LuckyCli::Generators::Validators::ProjectName.sanitize string
+end
+
+describe LuckyCli::Generators::Validators::ProjectName do
+  hyphenate = "sector-seven"
+  special_characters = "this string is emphatic"
+  title = "From the Earth to the Moon"
+
+  it "doesn't allow special characters" do
+    validate(special_characters).should be_false
+  end
+
+  it "corrects special characters" do
+    expected = "this_string_is_emphatic"
+    sanitize(special_characters).should eq expected
+  end
+
+  it "allows hyphens" do
+    validate(hyphenate).should be_true
+  end
+
+  it "doesn't allow uppercase characters" do
+    validate(title).should be_false
+  end
+
+  it "corrects uppercase characters" do
+    expected = "from_the_earth_to_the_moon"
+    sanitize(title).should eq expected
+  end
+end
--- a/src/generators/validators.cr
+++ b/src/generators/validators.cr
@@ -1,0 +1,11 @@
+module LuckyCli::Generators::Validators::ProjectName
+  extend self
+
+  def sanitize(name)
+    name.downcase.gsub(/[^a-z0-9_-]/, "_").strip('_')
+  end
+
+  def valid?(name)
+    sanitize(name) == name
+  end
+end
--- a/src/generators/web.cr
+++ b/src/generators/web.cr
@@ -10,8 +10,11 @@ class LuckyCli::Generators::Web
     project_name : String,
     @options : Options
   )
-    @project_dir = project_name.gsub(" ", "_")
-    @project_name = @project_dir.gsub("-", "_")
+    @project_dir = project_name
+    @project_name = project_name.gsub('-', '_')
+
+    validate_project_name @project_name
+
     @template_dir = File.join(__DIR__, "templates")
   end
 
@@ -177,6 +180,19 @@ class LuckyCli::Generators::Web
   private def ensure_directory_does_not_exist
     if Dir.exists?("./#{project_dir}")
       puts "Folder named #{project_dir} already exists, please use a different name".colorize.red.bold
+      exit
+    end
+  end
+
+  private def validate_project_name(name)
+    unless Validators::ProjectName.valid?(name)
+      message = <<-TEXT
+      Project name should only contain letters, numbers, underscores, and dashes.
+
+      How about: lucky init '#{Validators::ProjectName.sanitize(name)}'?
+      TEXT
+
+      puts message.colorize(:red)
       exit
     end
   end


src/generators/web.cr
```diff
@@ -10,8 +10,7 @@ class LuckyCli::Generators::Web
     project_name : String,
     @options : Options
   )
-    @project_dir = project_name.gsub(" ", "_")
-    @project_name = @project_dir.gsub("-", "_")
```
paulcsmith: The reason these were different is that it would keep the project directory as you typed it (for example `crystal-mastery`), but would use an underscore in the project_name, because that it what is used in the shard.yml and it will fail with a dash. I think that would still be nice to keep distinct so that you can have control over the project directory name.

robacarp:
Forgot to mention I also added a `shards build` target to make it quick and easy to build. Perhaps that should be in another pull?

shard.yml
```diff
@@ -6,6 +6,10 @@ authors:
 
 crystal: 0.25.0
 
+targets:
+  lucky:
+    main: src/lucky.cr
+
```
paulcsmith: Good idea! I think this is fine in this PR since it is a separate commit. I'll rebase & merge so it doesn't get squashed

paulcsmith:
Oh! And I had an idea for testing. It's not a full integration spec, but maybe having a `ProjectNameNormalizer` or `ProjectNameValidator` that just took a name and validate it or normalized it. Then you could test just that class in isolation

robacarp:
Building an entire class to normalize a string feels a little like overkill, but I could be persuaded. I like the idea of better error messages instead of magic.

Is the reason for converting spaces to underscores still a valid sanitization of the directory name?

paulcsmith:
I'm imaging if you did a validation class it would be a bit more complex because it would have to check if anything looked wrong and then it would need to raise an error, but that is up to you. I don't feel strongly about it. 

You could also consider doing a test like this: https://github.com/luckyframework/lucky_cli/blob/7c0e57355db315f0053521485ee1db0c6ece956f/spec/integration/init_web_spec.cr#L60

Where you try an invalid name and check that it output an error and check that no folder was created. 

As for converting spaces to underscores in folders, that might also be better as an error message. Something like "Spaces not allowed, please use a dash or underscore"

robacarp:
While I'm tinkering, should `[:alnum:]` be used instead of a-z for the benefits of internationalization? Would that require changes to the way the database name or the shards.yml are assembled?

paulcsmith:
Hmm good thought! I'm thinking that project names and shard names should not contain accent characters or anything special. It seems like that could cause issues. My other thought is that we can go with a more strict approach and relax it if it actually becomes a problem. My guess is that a bit stricter would be better

spec/integration/init_web_spec.cr
```diff
@@ -68,6 +68,28 @@ describe "Initializing a new web project" do
     message = "Folder named test-project already exists, please use a different name"
     output.to_s.strip.should contain(message)
   end
+
+  it "does not create project if project name is not a valid directory name" do
```
robacarp: I had a real hard time with this description, and I'm open to suggestions.

spec/lucky_cli/generators/validators_spec.cr
```diff
@@ -0,0 +1,53 @@
...
+      expected = "name_with_spaces"
+      LuckyCli::Generators::Validators::FolderName.sanitize(with_spaces).should eq expected
+    end
+
+    it "allows uppercase letters" do
+      LuckyCli::Generators::Validators::FolderName.valid?(uppercase).should be_true
+    end
+
+    it "doesnt correct uppercase" do
+      LuckyCli::Generators::Validators::FolderName.sanitize(uppercase).should eq uppercase
```
robacarp: These names are...verbose. I expected the describe to create a module around this so I could just use the `FolderName` part, but crystal couldn't resolve it.

src/generators/validators.cr
```diff
@@ -0,0 +1,23 @@
...
+
+      def sanitize(name)
+        {{ yield(name) }}
+      end
+
+      def valid?(name)
+        sanitize(name) == name
+      end
+    end
+  end
```
robacarp: Perhaps over-engineering?

src/generators/web.cr
```diff
@@ -10,8 +10,13 @@ class LuckyCli::Generators::Web
...
+    @project_dir = project_name
+    @project_name = project_name
+
+    validate(directory_name: @project_dir)
+    validate(project_name: @project_name)
```
robacarp: This is now not doing any manipulation of user input. If it's good it passes and if it's not it fails. I have a hunch that may not be the most helpful, but I'm not sure where to draw that line.

paulcsmith: What if instead of validating these separately, just the project name is validated. Since a valid project name will always be a valid directory name, that should cut down on the number of tests and the number of supporting lines of code

src/generators/web.cr
```diff
@@ -184,6 +189,29 @@ class LuckyCli::Generators::Web
     end
   end
 
+  private def validate(*, project_name)
+    unless Validators::ProjectName.valid?(project_name)
+      puts <<-TEXT
+      Project names should only contain letters, numbers, underscores, and dashes.
+
+      How about #{Validators::ProjectName.sanitize(project_name)}?
```
paulcsmith: I *love* the helpful suggestion here 👍

src/generators/web.cr
```diff
@@ -184,6 +189,29 @@ class LuckyCli::Generators::Web
...
+      TEXT
```
paulcsmith: Can you colorize this as red? `colorize.red`

src/generators/validators.cr
```diff
@@ -0,0 +1,23 @@
...
+
+  validate ProjectName do |name|
+    name.downcase.gsub(/[^a-z0-9_-]/,"_").strip('_')
+  end
+
+  validate FolderName do |name|
+    name.gsub(/[^a-zA-Z0-9_-]/,"_").strip('_')
+  end
+end
```
paulcsmith: If it is just a project name validator then this can probably be done without macros, and instead some class methods on a class or maybe a module with `extend self`?

spec/lucky_cli/generators/validators_spec.cr
paulcsmith: In situations like this I usually extract a private method in the spec so it is easier to call and easier to rename later. Something like `private def sanitize(value)`

spec/integration/init_web_spec.cr
```diff
@@ -68,6 +68,28 @@ describe "Initializing a new web project" do
...
+    output.to_s.strip.should contain(message)
+  end
+
+  it "does not create project if project name is invalid" do
+    output = IO::Memory.new
+    Process.run(
+      "crystal src/lucky.cr init 'test *roject'",
+      output: output,
+      shell: true
+    )
```
paulcsmith: I added a module that might help with this https://github.com/luckyframework/lucky_cli/blob/master/spec/support/should_run_successfully.cr

So I think you could do `should_run_successfully "crystal src/lucky.cr init 'test *roject'"`, io: output`

spec/lucky_cli/generators/validators_spec.cr
```diff
@@ -0,0 +1,40 @@
...
+
+private def validate(string)
+  LuckyCli::Generators::Validators::ProjectName.valid? string
+end
+
+private def sanitize(string)
+  LuckyCli::Generators::Validators::ProjectName.sanitize string
+end
+
+
```
paulcsmith: Could you run `crystal tool format spec src`? I think that should clean up this extra newline

spec/lucky_cli/generators/validators_spec.cr
```diff
@@ -0,0 +1,40 @@
...
+describe LuckyCli::Generators::Validators do
+  context "ProjectName" do
+    hyphenate = "sector-seven"
+    special_characters = "this string is emphatic"
+    title = "From the Earth to the Moon"
+
+    it "doesnt allow special characters" do
```
paulcsmith: doesnt is missing an apostrophe

spec/lucky_cli/generators/validators_spec.cr
```diff
@@ -0,0 +1,40 @@
...
+  LuckyCli::Generators::Validators::ProjectName.valid? string
+end
+
+private def sanitize(string)
+  LuckyCli::Generators::Validators::ProjectName.sanitize string
+end
+
+
+describe LuckyCli::Generators::Validators do
+  context "ProjectName" do
```
paulcsmith: Rather than adding a context, I think this could be moved into the describe. That way it is a bit more inline with the rest of the specs. What do you think?

src/generators/validators.cr
```diff
@@ -0,0 +1,13 @@
+module LuckyCli::Generators::Validators
+  module ProjectName
+    extend self
```
paulcsmith: Noice. I love this pattern

src/generators/validators.cr
```diff
@@ -0,0 +1,13 @@
...
+
+    def sanitize(name)
+      name.downcase.gsub(/[^a-z0-9_-]/,"_").strip('_')
+    end
+
+    def valid?(name)
+      sanitize(name) == name
```
paulcsmith: Nice and simple. Love it

src/generators/validators.cr
```diff
@@ -0,0 +1,13 @@
+module LuckyCli::Generators::Validators
+  module ProjectName
```
paulcsmith: Could you put this on the line above? It's a small thing but it'd be more in line with the other Lucky classes/modules

spec/lucky_cli/generators/validators_spec.cr
```diff
@@ -0,0 +1,40 @@
...
+    end
+
+    it "doesnt allow uppercase characters" do
+      validate(title).should be_false
+    end
+
+    it "corrects uppercase characters" do
+      expected = "from_the_earth_to_the_moon"
+      sanitize(title).should eq expected
+    end
```
paulcsmith: The tests read quite nicely!

src/generators/web.cr
```diff
@@ -10,8 +10,11 @@ class LuckyCli::Generators::Web
     project_name : String,
     @options : Options
   )
-    @project_dir = project_name.gsub(" ", "_")
-    @project_name = @project_dir.gsub("-", "_")
+    @project_dir = project_name
+    @project_name = project_name
```
paulcsmith: It looks like the sanitizer/validator allows dashes (which is a expected), but the dashes should be changed to "_". Otherwise the shard name and generate file will have a dash and will break deployment on Heroku because it will not be able to find the file (it'll be looking for one with an underscore). I think using what was there previously would fix it. `@project_name = @project_dir.gsub("-", "_")`

paulcsmith:
Thanks for your help with this 👍

PR #262, opened 2018-11-01
Adding 0 byte default favicon to generated web app

jwoertink:
This fixes https://github.com/luckyframework/lucky/issues/646

Right now a default generated lucky app will throw 500s in development when the browser auto-requests the favicon (404 in production). This just adds a blank default one so the file exists. The developer can change it out when they want.


paulcsmith:
When the browser requests this does it show an error because it is zero bytes? Or does it just show nothing? I haven't really done anything with favicon so I'm not really sure how it is handled

edwardloveall:
I just did some experiments and it looks like it just shows nothing. Safari is even smart enough to not allocate space for the icon next to the page title in the tab bar if it's empty.

paulcsmith:
Thanks for checking @edwardloveall! I’ll go ahead and merge this in!

paulcsmith:
Thanks @jwoertink 🙏

PR #264, opened 2018-11-02
Rename banner to summary for tasks. fixes #115

### LINKED ISSUE #115: Change banner -> summary
I think `banner` isn't a good name. `description` or `summary` might be better

jwoertink:
The only spot left that mentions `banner` is the opt parser which actually is called banner. This will technically be a breaking change, so it should probably be noted somewhere for the next version of lucky.


src/lucky_cli/task.cr
```diff
@@ -7,9 +7,9 @@ abstract class LuckyCli::Task
...
   end
 
-  macro banner(banner_text)
-    def banner
-      {{banner_text}}
+  macro summary(summary_text)
+    def summary
+      {{summary_text}}
     end
   end
```
paulcsmith: Could you add another macro for `banner` that just raises and says to use `summary` instead? Something like this:

```crystal
macro banner
  {% raise "'banner' has been renamed to 'summary'" %}
end
```

PR #307, opened 2019-02-28
Make it clear backdoor is in test only

paulcsmith:
It was scary to see "sign_in_through_backdoor" in your logs.

This name change makes things shorter and also clarifies that the
backdoor is for testing only.


paulcsmith:
Thanks for reviewing!

PR #311, opened 2019-03-01
Added 2 methods to app which will become abstract methods in BaseApp

jwoertink:
Per [this conversation](https://github.com/luckyframework/lucky/issues/707), these two methods will become abstract on the parent class. This allows the dev to quickly customize if need be once the app is generated. 

Also, I added the 3rd argument of false since false is already the default. This gives a quick insight in to being able to set it to true. I think I want to add a comment on here too, but wording was weird. 

> The third argument disabled the SO_REUSEPORT option

If you have an idea, let me know.

### DIFF
--- a/src/web_app_skeleton/src/app.cr.ecr
+++ b/src/web_app_skeleton/src/app.cr.ecr
@@ -50,4 +50,13 @@ class App < Lucky::BaseApp
       Lucky::RouteNotFoundHandler.new,
     ]
   end
+
+  def protocol
+    "http"
+  end
+
+  def listen
+    server.bind_tcp(host, port, false)
+    server.listen
+  end
 end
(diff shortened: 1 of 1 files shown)


src/web_app_skeleton/src/app.cr.ecr
```diff
@@ -50,4 +50,13 @@ class App < Lucky::BaseApp
       Lucky::RouteNotFoundHandler.new,
     ]
   end
+
+  def protocol
+    "http"
+  end
+
+  def listen
+    server.bind_tcp(host, port, false)
```
paulcsmith: I love the idea of showing people they can use the reuse port option. To make it a bit clearer what if we added a keyword arg: `server.bind_tcp(host, port, reuse_port: false)`. Maybe the comment could then be a link to the `bind_tcp` docs in Crystal

jwoertink: ah yup. Good call.

PR #312, opened 2019-03-01
Remove unnecessary Authentic abstractions

paulcsmith:
Implementing some ideas discussed with @robacarp 

Still more to come, but wanted to get this one out there

### DIFF
--- a/src/base_authentication_app_skeleton/src/actions/mixins/auth/allow_guests.cr
+++ b/src/base_authentication_app_skeleton/src/actions/mixins/auth/allow_guests.cr
@@ -1,4 +1,4 @@
-module Auth::SkipRequireSignIn
+module Auth::AllowGuests
   macro included
     skip require_sign_in
   end
--- a/src/base_authentication_app_skeleton/src/actions/mixins/auth/redirect_signed_in_users.cr
+++ b/src/base_authentication_app_skeleton/src/actions/mixins/auth/redirect_signed_in_users.cr
@@ -1,10 +1,10 @@
-module Auth::RedirectIfSignedIn
+module Auth::RedirectSignedInUsers
   macro included
     include Auth::AllowGuests
-    before redirect_if_signed_in
+    before redirect_signed_in_users
   end
 
-  private def redirect_if_signed_in
+  private def redirect_signed_in_users
     if current_user?
       flash.success = "You are already signed in"
       redirect to: Home::Index
@@ -13,7 +13,7 @@ module Auth::RedirectIfSignedIn
     end
   end
 
-  # current_user returns nil so that it won't accidentally be used
+  # current_user returns nil because signed in users are redirected.
   def current_user
   end
 end
--- a/src/base_authentication_app_skeleton/src/forms/mixins/user_from_email.cr
+++ b/src/base_authentication_app_skeleton/src/forms/mixins/user_from_email.cr
@@ -1,5 +1,5 @@
-module FindAuthenticatable
-  private def find_authenticatable
+module UserFromEmail
+  private def user_from_email : User?
     email.value.try do |value|
       UserQuery.new.email(value).first?
     end
--- a/src/base_authentication_app_skeleton/src/forms/password_reset_request_form.cr
+++ b/src/base_authentication_app_skeleton/src/forms/password_reset_request_form.cr
@@ -1,8 +1,20 @@
 class PasswordResetRequestForm < Avram::VirtualForm
-  include Authentic::FormHelpers
-  include FindAuthenticatable
+  # You can modify this in src/forms/mixins/user_from_email.cr
+  include UserFromEmail
 
   virtual email : String
+
+  # Run validations and yield the form and the user if valid
+  def submit
+    user = user_from_email
+    validate(user)
+
+    if valid?
+      yield self, user
+    else
+      yield self, nil
+    end
+  end
 
   def validate(user : User?)
     validate_required email
--- a/src/base_authentication_app_skeleton/src/forms/sign_in_form.cr
+++ b/src/base_authentication_app_skeleton/src/forms/sign_in_form.cr
@@ -1,19 +1,30 @@
 class SignInForm < Avram::VirtualForm
-  include Authentic::FormHelpers
-  include FindAuthenticatable
+  # You can modify this in src/forms/mixins/user_from_email.cr
+  include UserFromEmail
 
   virtual email : String
   virtual password : String
 
-  # This method is called to allow you to determine if a user can sign in.
-  # By default it validates that the user exists and the password is correct.
+  # Run validations and yields the form and the user if valid
+  def submit
+    user = user_from_email
+    validate_credentials(user)
+
+    if valid?
+      yield self, user
+    else
+      yield self, nil
+    end
+  end
+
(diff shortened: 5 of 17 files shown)


src/base_authentication_app_skeleton/src/forms/mixins/user_from_email.cr
```diff
@@ -1,5 +1,5 @@
-module FindAuthenticatable
-  private def find_authenticatable
+module UserFromEmail
```
paulcsmith: Clarify what exactly this module is doing

src/base_authentication_app_skeleton/src/actions/mixins/auth/allow_guests.cr
```diff
@@ -1,4 +1,4 @@
-module Auth::SkipRequireSignIn
+module Auth::AllowGuests
```
paulcsmith: Rather than saying what it is skipping, I'm thinking it is better to say what it should be used for, allowing guests.

src/base_authentication_app_skeleton/src/forms/password_reset_request_form.cr
```diff
@@ -1,9 +1,21 @@
...
+  def submit
+    user = user_from_email
+    validate(user)
+
+    if valid?
+      yield self, user
+    else
+      yield self, nil
+    end
+  end
```
paulcsmith: Moved here so it is easier to see what is going on, and how the form works. Thoughts?

src/base_authentication_app_skeleton/src/actions/mixins/auth/redirect_signed_in_users.cr
```diff
@@ -0,0 +1,19 @@
+module Auth::RedirectSignedInUsers
```
paulcsmith: Renamed. Used to be `Auth::RedirectIfSignedIn`

paulcsmith:
@edwardloveall Would love to hear your thoughts on these changes :D

src/base_authentication_app_skeleton/src/forms/password_reset_request_form.cr
edwardloveall: Yup, I think having this abstraction live here is better than in a library. This feels like the tradeoffs that [Oath] makes where most of the user-level logic lives in the app itself doesn't force you into any particular pattern. I like that kind of tradeoff.

[Oath]: https://github.com/halogenandtoast/oath

paulcsmith: Is this clearer?

```crystal
  def submit
    user = user_from_email
    validate_credentials(user)
    # set user to `nil` instead of using an if with 2 yields
    user = nil unless valid?
    yield self, user
  end
```

src/base_authentication_app_skeleton/src/forms/sign_in_form.cr
```diff
@@ -1,24 +1,38 @@
...
   #    end
-  private def validate(user : User?)
+  private def validate_credentials(user)
     if user
       unless Authentic.correct_password?(user, password.value.to_s)
         password.add_error "is wrong"
       end
     else
+      # Usually ok to say that an email is not in the system:
+      # https://kev.inburke.com/kevin/invalid-username-or-password-useless/
```
edwardloveall: This is a good article. I like the takeaways he recommends. 👍

src/browser_authentication_app_skeleton/src/actions/sign_ups/new.cr
```diff
@@ -1,5 +1,5 @@
 class SignUps::New < BrowserAction
-  include Auth::RedirectIfSignedIn
+  include Auth::RedirectSignedInUsers
```
edwardloveall: With all these renames, would it makes sense to put this in a higher level base controller? I know inheritance over compostability is heresy these days, but I feel like it makes sense to make some kind of `AuthenticatedAction` since most (but not all) of the app would use it.

edwardloveall: Or actually... can this just be the default in the browser action and instead you include `Auth::AllowGuests` to disable that behavior in the few cases that you need? Maybe that module would stub out the methods that prevent someone from logging in. Just thinking out loud here. I'm sure you've considered this much longer than I have 🙂

src/base_authentication_app_skeleton/src/forms/password_reset_request_form.cr
paulcsmith: @edwardloveall I just added another comment here. Thoughts?

edwardloveall: I think the first one is clearer, although this is a bit prettier. This makes it slightly more difficult to keep track of what state `user` is in.

src/browser_authentication_app_skeleton/src/actions/sign_ups/new.cr
paulcsmith: @edwardloveall You can't see the change since it wasn't renamed, but in `BrowserAction` there is `Auth::RequireSignIn` by default. You only include these other ones when you want to change the default behavior which requires sign in. Is that kind of what you mean?

src/base_authentication_app_skeleton/src/forms/password_reset_request_form.cr
paulcsmith: I agree. I want to find a simpler way (I had an idea for something in 
https://github.com/luckyframework/avram/issues/69) but I think you're right that the conditionals make it clearer 👍

src/browser_authentication_app_skeleton/src/actions/sign_ups/create.cr
```diff
@@ -1,9 +1,9 @@
 class SignUps::Create < BrowserAction
-  include Auth::RedirectIfSignedIn
+  include Auth::RedirectSignedInUsers
 
   route do
     SignUpForm.create(params) do |form, user|
-      if user
+      if form.valid? && user
```
paulcsmith: Here's another idea. Always check if the form is valid first. Then we can just do `yield self, user` without the conditional. I really haven't thought this through a ton. I'm not totally sure I like it, but thought I'd see what others think.

One downside if you now *always* have to check if the form is valid, and if you forgot and only check for user you might accidentally sign a user in! :S

src/browser_authentication_app_skeleton/src/actions/sign_ups/new.cr
edwardloveall: I think I'm both under- and over-thinking it. I'll look at this again tomorrow and give you some better feedback 😴

edwardloveall: Ok, with fresh eyes I think this was just a silly mistake on my part. I was reading `RedirectSignedInUsers` as something that required users to sign in. It didn't make sense to me to include that functionality everywhere individually instead of having it exist in a base action.

And I think it does make sense to separate that from the AllowGuests mixin. Sorry! Disregard.

paulcsmith:
Thank you for the review @edwardloveall! If anyone else stumbles on this and has feedback, feel free to leave it and we can open a new PR

PR #315, opened 2019-03-03
Move methods to class components

paulcsmith:
* Easier to find where a component is defined
* Works with go to definition
* Doesn't add a bunch of methods to all the page by including them
* Allows namespacing , e.g. `Users::FormFields`
* Does not require including modules. Can mount from just about
  anywhere.
* Full power of classes


src/browser_app_skeleton/src/components/shared/field.cr
```diff
@@ -1,34 +1,34 @@
...
+    label_for @field
 
-    unless hide_errors
-      errors_for field
+    with_defaults field: @field do |html|
+      yield html
     end
+
+    mount Shared::FieldErrors, @field
   end
```
paulcsmith: I still need to figure out a good way to hide labels and errors. I'm thinking maybe it can be another `needs` with a default, or I may change the `mount` method to use `.new` so you can call methods on the component before rendering: `mount Shared::Field.new(form.email).hide_label.hide_errors, &.email_input`

src/browser_authentication_app_skeleton/src/pages/auth_layout.cr
```diff
@@ -1,15 +1,25 @@
...
-  include Shared::Layout
+  include Lucky::HTMLPage
+
+  abstract def content
+  abstract def page_title
+
+  # The default page title. It is passed to `Shared::LayoutHead`.
+  #
+  # Add a `page_title` method to pages to override it. You can also remove
+  # This method so every page is required to have its own page title.
```
paulcsmith: Also tried to clarify the page_title help text

src/browser_app_skeleton/src/components/shared/field.cr
```diff
@@ -1,34 +1,34 @@
...
+class Shared::Field(T) < BaseComponent
+  needs field : Avram::FillableField(T)
 
-    yield field
+  def render
+    label_for @field
 
-    unless hide_errors
-      errors_for field
+    with_defaults field: @field do |html|
```
paulcsmith: https://github.com/luckyframework/lucky/pull/713

paulcsmith: I wonder if calling the yield variable `tag` would make more sense:

```
with_defaults field: @field do |tag|
  yield tag
end
```

src/browser_app_skeleton/src/components/shared/field.cr
HarrisonB: I like the extensibility of this idea. If we’re moving towards objects over just methods, why not give the full power of objects?

paulcsmith: Yeah that’s kinda what I’m thinking to. Thanks for the feedback :) I think I’ll make that change in Lucky!

src/browser_app_skeleton/src/components/shared/field.cr
```diff
@@ -1,34 +1,47 @@
...
 
-    unless hide_errors
-      errors_for field
+    # You can add more default options here. For example:
+    #
+    #    with_defaults field: @field, class: "input"
+    #
+    # Will add the class "input" to the generated HTML.
+    with_defaults field: @field do |tag_builder|
+      yield tag_builder
```
paulcsmith: Renamed to `tag_builder`. I *think* that's clearer

src/browser_authentication_app_skeleton/src/pages/auth_layout.cr
```diff
@@ -1,15 +1,25 @@
...
   def render
     html_doctype
 
     html lang: "en" do
-      shared_layout_head
+      mount Shared::LayoutHead.new(page_title: page_title, context: @context)
 
       body do
-        render_flash
+        mount Shared::FlashMessages.new(@context.flash)
```
paulcsmith: Uses `new` so you can do all kinds of stuff with the class before it is rendered/mounted

PR #322, opened 2019-03-13
disable colorize output for production by default. 

### LINKED ISSUE #317: Adding option to disable colorize
Crystal has an [option to disable colorize](https://crystal-lang.org/api/0.27.2/Colorize.html#enabled%3D%28enabled%3ABool%29-class-method). I ran in to an issue recently where errors were shipped off to bugsnag, and pretty much unreadable.

screen shot 2019-03-05 at 11 30 58 am

Granted, this is from Clear, and not Avram, but the concept is still similar. In a production setting, coloring usually doesn't matter. You can ship your logs off to 3rd party services like timber, or loggly, and some of those don't parse the coloring sequences. 

This would be a good option for us to add something like a `config/color.cr` or whatever.

```crystal
Colorize.enabled = Lucky::Env.development?
```

jwoertink:
Fixes #317 

I'm wondering if this might go better in the logger.cr file that already exists since this is more related to logging?

### DIFF
--- a/src/web_app_skeleton/config/color.cr
+++ b/src/web_app_skeleton/config/color.cr
@@ -1,0 +1,4 @@
+# This enables the color output when in development
+# Check out the Colorize docs for more information
+# https://crystal-lang.org/api/Colorize.html
+Colorize.enabled = Lucky::Env.development?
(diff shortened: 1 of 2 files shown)


paulcsmith:
Good idea! Yeah I think putting it in logger makes sense 👍

PR #323, opened 2019-03-13
When in api_only, skip rendering exception page on errors. 

### LINKED ISSUE #321: Don't render exception page if requesting JSON

jwoertink:
Fixes #321 

I took a stab at this. Not sure if this is what you were going for, but it made sense to me. Skip calling the render exception page if it's api_only.


paulcsmith:
I'm gonna close this in favor of #395 which only render the exception page if it is HTML. So even if you have an HTML and API app it will still no render the HTML page if you request JSON!

PR #325, opened 2019-03-16
Add building from source instructions

volodymyrshevhuk:
Had some problems with Lucky CLI installation by linux version of homebrew(formerly linuxbrew)
Explicit instructions about building from source could be useful


jwoertink:
I think these instructions are great. I've had to build from source a few times so I could use some lucky features not released yet.

PR #333, opened 2019-04-19
v0.14.0 and upgrade to Crystal 0.28.0

paulcsmith:


jwoertink:
This probably just needs a rebase from master and then it'll be good. I didn't see any other changes that would affect for crystal 0.28.0. The next version of crystal is being released soon, so we can start prepping for that once this is merged in and locked.

paulcsmith:
Oh I was supposed to close this! This isn't actually based on master so it should live as a permanent branch. It is actually based on v0.13.2 so can't really be rebased onto master

jwoertink:
ah. Gotcha.

PR #337, opened 2019-04-26
Adding some help comments for setting up a DB

jwoertink:
This is a follow up to https://github.com/luckyframework/avram/pull/82#issuecomment-487142573

In some cases, someone may not have their default pg username as "postgres", so offer up a few suggestions as alternatives. Also, just in case they don't need a password, be clear that a empty string `""` is fine for that.

### DIFF
--- a/src/web_app_skeleton/config/database.cr.ecr
+++ b/src/web_app_skeleton/config/database.cr.ecr
@@ -7,7 +7,9 @@ Avram::Repo.configure do |settings|
     settings.url = ENV["DATABASE_URL"]? || Avram::PostgresURL.build(
       database: database_name,
       hostname: ENV["DB_HOST"]? || "localhost",
+      # Common default usernames may be "postgres", "root", or your system username
       username: ENV["DB_USERNAME"]? || "postgres",
+      # If no password is necessary, use an empty string ""
       password: ENV["DB_PASSWORD"]? || "postgres"
     )
   end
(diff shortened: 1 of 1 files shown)


src/web_app_skeleton/config/database.cr.ecr
```diff
@@ -7,7 +7,9 @@ Avram::Repo.configure do |settings|
     settings.url = ENV["DATABASE_URL"]? || Avram::PostgresURL.build(
       database: database_name,
       hostname: ENV["DB_HOST"]? || "localhost",
+      # Common default usernames may be "postgres", "root", or your system username
```
paulcsmith: I think if we say "common default usernames" we can remove "may be". Might be helpful to tell users who to get system username: `Common default usernames are "postgres", "root", or your system username (run 'whoami')`

Thoughts on that?

src/web_app_skeleton/config/database.cr.ecr
```diff
@@ -7,7 +7,9 @@ Avram::Repo.configure do |settings|
...
       username: ENV["DB_USERNAME"]? || "postgres",
+      # If no password is necessary, use an empty string ""
```
paulcsmith: Might be worth saying that sometimes default Postgres credentials require no password?

```
# Some Postgres installations require no password. Use "" if that is the case.
```

Thoughts on that? Does it make sense

PR #342, opened 2019-05-31
Allow for setting a custom help message on tasks. 

### LINKED ISSUE #340: Extra help for tasks
When you run `lucky -h`, you get output like:

```
Usage: lucky name.of.task [options]

Available tasks:

  ▸ build.release             Compile app for production
  ▸ db.create                 Create the database
  ▸ db.create_required_seeds  Add database records required for the app to work
  ▸ db.create_sample_seeds    Add sample database records helpful for development
  ▸ db.drop                   Drop the database
  ▸ db.migrate                Migrate the database
  ▸ db.migrate.one            Run the next pending migration
  ▸ db.redo                   Rollback then run the last migration
  ▸ db.rollback               Rollback the last migration
  ▸ db.rollback_all           Rollback all migrations
  ▸ exec                      Execute code. Use this in place of a console/REPL
  ▸ gen.action.api            Generate a new api action
  ▸ gen.action.browser        Generate a new browser action
  ▸ gen.component             Generate a new component
  ▸ gen.migration             Generate a new migration
  ▸ gen.model                 Generate a model, query, and form
  ▸ gen.page                  Generate a new page
  ▸ gen.resource.browser      Generate a resource (model, form, query, actions, and pages)
  ▸ gen.secret_key            Generate a new secret key
  ▸ routes                    Show all the routes for the app
  ▸ watch                     Start and recompile project when files change
```

For someone brand new to lucky, this is a little daunting, as well as slightly confusing.  You know that you can run a migration like `lucky gen.migration`, and you know it takes options like `[options]`, but you don't know **what** that actually looks like. _What_ `[options]` can you pass? 

For this case, it would be nice if you could do `lucky -h gen.migration` and then see output similar to:

```
Usage: lucky gen.migration AddColumnToModel

Example:

  ▸ lucky gen.migration AddNameToUsers     
```

Of course this would mean that CLI would need an optional `example` macro so your custom tasks, or tasks created by shards could be looped in.

jwoertink:
fixes #340

This PR adds in the ability to set a custom `help_message` method on tasks so people can pass one of `-h`, `--help`, or just `help` to a task to see a more detailed help message on how to run that specific task

### DIFF
--- a/spec/lucky_cli/runner_spec.cr
+++ b/spec/lucky_cli/runner_spec.cr
@@ -21,6 +21,30 @@ describe LuckyCli::Runner do
       .should have_called_my_cool_task
   end
 
+  it "prints the help_message for a found task when a help flag is passed" do
+    LuckyCli::Runner
+      .run(args: ["my.cool_task", "--help"])
+      .should have_default_help_message
+    LuckyCli::Runner
+      .run(args: ["my.cool_task", "-h"])
+      .should have_default_help_message
+    LuckyCli::Runner
+      .run(args: ["my.cool_task", "help"])
+      .should have_default_help_message
+  end
+
+  it "prints a custom help_message when set" do
+    LuckyCli::Runner
+      .run(args: ["my.custom_name", "-h"])
+      .should eq "Run with 'lucky my.custom_name'"
+  end
+
+  it "still calls a found task with other options passed" do
+    LuckyCli::Runner
+      .run(args: ["my.cool_task", "Taco", "--with-guac"])
+      .should have_called_my_cool_task
+  end
+
   it "does not call the task if no args passed" do
     LuckyCli::Runner
       .run(args: [] of String)
@@ -42,4 +66,8 @@ end
 
 private def have_called_my_cool_task
   eq :my_cool_task_was_called
+end
+
+private def have_default_help_message
+  eq "Run this task with `lucky my.cool_task`"
 end
--- a/spec/lucky_cli/runner_spec.cr
+++ b/spec/lucky_cli/runner_spec.cr
@@ -22,24 +22,20 @@ describe LuckyCli::Runner do
   end
 
   it "prints the help_message for a found task when a help flag is passed" do
-    LuckyCli::Runner
-      .run(args: ["my.cool_task", "--help"])
-      .should have_default_help_message
-    LuckyCli::Runner
-      .run(args: ["my.cool_task", "-h"])
-      .should have_default_help_message
-    LuckyCli::Runner
-      .run(args: ["my.cool_task", "help"])
-      .should have_default_help_message
+    %w(--help -h help).each do |help_arg|
+      io = IO::Memory.new
+      LuckyCli::Runner.run(args: ["my.cool_task", help_arg], io: io)
+      io.to_s.chomp.should have_default_help_message
+    end
   end
 
   it "prints a custom help_message when set" do
-    LuckyCli::Runner
-      .run(args: ["my.custom_name", "-h"])
-      .should eq "Run with 'lucky my.custom_name'"
+    io = IO::Memory.new
+    LuckyCli::Runner.run(args: ["my.custom_name", "-h"], io: io)
+    io.to_s.chomp.should eq "Custom help message"
   end
 
-  it "still calls a found task with other options passed" do
+  it "still calls a found task with non-help options" do
     LuckyCli::Runner
       .run(args: ["my.cool_task", "Taco", "--with-guac"])
       .should have_called_my_cool_task
@@ -69,5 +65,5 @@ private def have_called_my_cool_task
 end
 
 private def have_default_help_message
-  eq "Run this task with `lucky my.cool_task`"
+  eq "Run this task with 'lucky my.cool_task'"
 end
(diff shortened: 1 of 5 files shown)


spec/lucky_cli/runner_spec.cr
```diff
@@ -43,3 +67,7 @@ end
 private def have_called_my_cool_task
   eq :my_cool_task_was_called
 end
+
+private def have_default_help_message
```
paulcsmith: I love the helper method here 👍

spec/lucky_cli/runner_spec.cr
```diff
@@ -21,6 +21,30 @@ describe LuckyCli::Runner do
...
+  it "prints the help_message for a found task when a help flag is passed" do
+    LuckyCli::Runner
+      .run(args: ["my.cool_task", "--help"])
+      .should have_default_help_message
+    LuckyCli::Runner
+      .run(args: ["my.cool_task", "-h"])
+      .should have_default_help_message
+    LuckyCli::Runner
+      .run(args: ["my.cool_task", "help"])
+      .should have_default_help_message
```
paulcsmith: Maybe to make it clear what each of these tests is doing you could iterate over an array:

```crystal
%w( -h --help help).each do |help_arg|
  LuckyCli::Runner
     # ... run it and assert it
end
```

spec/lucky_cli/runner_spec.cr
```diff
@@ -21,6 +21,30 @@ describe LuckyCli::Runner do
...
+  end
+
+  it "prints a custom help_message when set" do
+    LuckyCli::Runner
+      .run(args: ["my.custom_name", "-h"])
+      .should eq "Run with 'lucky my.custom_name'"
+  end
+
+  it "still calls a found task with other options passed" do
```
paulcsmith: Rather than "with other options passed" I think it would be extra clear if it were something like `with non help options`. What do you think?

spec/support/tasks.cr
```diff
@@ -10,6 +10,10 @@ class Some::Other::Task < LuckyCli::Task
   summary "bar"
   name "my.custom_name"
 
+  def help_message
+    "Run with 'lucky my.custom_name'"
```
paulcsmith: Maybe "Custom help message" just so it is extra clear the is not the default one?

src/lucky_cli/task.cr
```diff
@@ -7,6 +7,19 @@ abstract class LuckyCli::Task
     def name
       "{{@type.name.gsub(/::/, ".").underscore}}"
     end
+
+    def help_message
+      "Run this task with `lucky #{name}`"
```
paulcsmith: I think we tend to use single quotes instead of backpacks in most places in the help messages in Lucky

src/lucky_cli/task.cr
```diff
@@ -7,6 +7,19 @@ abstract class LuckyCli::Task
...
+    end
+
+    def print_help_or_call(args : Array(String))
+      if args.any? { |arg| ["--help", "-h", "help"].includes?(arg) }
```
paulcsmith: I think it would be good to do something like this `args.size == 1 && %w(-h --help help).includes(args.first)`. That way it only works with the first arg. Also might be worth extracting a method like `wants_help_message?(args)` or something like that

src/lucky_cli/task.cr
```diff
@@ -7,6 +7,19 @@ abstract class LuckyCli::Task
...
+        STDERR.puts help_message
+        help_message
```
paulcsmith: I think this can be removed  and the return type set to `Nil` so the result of this call is not used. That may mean reworking the tests to pass in an optional `io` argument.

```crystal
def print_help_or_call(args : Array(String), io : IO = STDERR) : Nil
  io.puts help_message
end

# in the tests
io = IO::Memory.new

LuckyCli::Runner.new
  .print_help_or_call(args, io: io)
io.to_s.should have_helper_message
```

src/lucky_cli/task.cr
jwoertink: This one is a little tricky because technically if `args.first` is `--help`, then it prints the main help with `lucky --help`. So it might have to be something like `args.size == 2` or `args.size > 1` and `[].includes? args[1]`

paulcsmith: Ohhh I see. Let's stick with what you've got then. I have some ideas for making this whole thing more robust in the future. I did it differently in `up` and it is a lot nicer. Basically duplicates the args and removes the first arg once it finds a matching command

src/lucky_cli/task.cr
jwoertink: oh that's interesting. I can dig it!

spec/lucky_cli/runner_spec.cr
```diff
@@ -21,6 +21,26 @@ describe LuckyCli::Runner do
...
+
+  it "prints a custom help_message when set" do
+    io = IO::Memory.new
+    LuckyCli::Runner.run(args: ["my.custom_name", "-h"], io: io)
+    io.to_s.chomp.should eq "Custom help message"
+  end
+
+  it "still calls a found task with non-help options" do
+    LuckyCli::Runner
+      .run(args: ["my.cool_task", "Taco", "--with-guac"])
```
paulcsmith: Tacos!

PR #348, opened 2019-06-15
Rename Form -> SaveOperation

paulcsmith:
Part of https://github.com/luckyframework/avram/pull/104

PR #349, opened 2019-06-18
Use 'SaveOperation', 'permit_columns', and 'PermittedAttribute'

smadeja:
I guess this will be needed after #348 is merged. It reflects the changes introduced in https://github.com/luckyframework/avram/pull/105/commits/23c4158e1dd281b102b847f4f2148d5ba988c87c.


paulcsmith:
Thank you so much! Could you also change the spots where we have `FillableField` to `PermittedAttribute`?

PR #354, opened 2019-06-22
Skip logging static assets

paulcsmith:
See https://github.com/luckyframework/lucky/pull/824


paulcsmith:
cc @jwoertink @edwardloveall if you're using Lucky master you'll want to do something like this in your apps to make them work

paulcsmith:
Just realized we need to make sure that it still logs 404s....

PR #362, opened 2019-07-10
script/setup uses /bin/bash which is not available everywhere

peterhoeg:
Some distributions do not have ```/bin/bash``` so use ```/usr/bin/env``` to find it.

### DIFF
--- a/script/test
+++ b/script/test
@@ -1,4 +1,4 @@
-#! /bin/bash
+#!/usr/bin/env bash
 
 # Exit if any subcommand fails
 set -e
--- a/script/setup
+++ b/script/setup
@@ -1,4 +1,4 @@
-#! /bin/bash
+#!/usr/bin/env bash
 
 # Exit if any subcommand fails
 set -e
--- a/src/web_app_skeleton/script/setup.ecr
+++ b/src/web_app_skeleton/script/setup.ecr
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Exit if any subcommand fails
 set -e


jwoertink:
I wonder if this should actually change to `sh` like what crystal did [here](https://github.com/crystal-lang/crystal/pull/3809), and [here](https://github.com/crystal-lang/crystal/pull/5468)

The main points from [this SO post](https://stackoverflow.com/questions/5725296/difference-between-sh-and-bash)

```
Which one to use

For my own scripts, I prefer sh for the following reasons:

    it is standardized
    it is much simpler and easier to learn
    it is portable across POSIX systems — even if they happen not to have bash, they are required to have sh
```

@paulcsmith what are your thoughts on this?

paulcsmith:
Thanks for starting this up @peterhoeg!

@jwoertink yeah I think sh makes a lot of sense. @peterhoeg could  you update to use sh as described in the Jeremy's comment?

peterhoeg:
>   @jwoertink yeah I think sh makes a lot of sense. @peterhoeg could you
>   update to use sh as described in the Jeremy's comment?

No problem - I haven't looked yet if there are any bashisms in those scripts but should be straight forward.

script/test
```diff
@@ -1,4 +1,4 @@
-#! /bin/bash
+#!/bin/sh
```
jwoertink: line 20 of this file needs to change

```diff
- function finish {
+ finish() {
   docker-compose down
}
```

up
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
```
jwoertink: line 19:

```diff
- if [ "$MACHINE" == "UNKNOWN" ]; then
+ if [ "$MACHINE" = "UNKNOWN" ]; then
```

line 34:

```diff
- if [ "$1" == "ssh" ]; then
+ if [ "$1" = "ssh" ]; then
```

jwoertink:
@peterhoeg after a bit of discussion with @paulcsmith we decided to use your original proposed solution with the env for bash as opposed to using `sh`. The main problem is that `sh` doesn't support the `-o pipefail` option natively, and since these scripts here use pipes a ton, that can lead to some missed errors. Using your original proposed change will open us up for compatibility a little more, but still keep things working how we want. We can revisit moving to full POSIX compat later once it starts coming up more. 

Thanks!

peterhoeg:
Fixed.

peterhoeg:
I squashed the commits - hope that's ok.

jwoertink:
Yup, that's cool. Looks like the CI is happy, and this change looks good to me. Thanks!

PR #364, opened 2019-07-10
generate a .env during script/setup if one does not exist

### LINKED ISSUE #359: script/setup should populate blank .env if none exists
So you don't see this when clinging and setting up someone else's project

Screen Shot 2019-07-02 at 2 16 13 PM

jwoertink:
This fixes #359 

Any idea where a spec would go for this?

### DIFF
--- a/spec/integration/init_web_spec.cr
+++ b/spec/integration/init_web_spec.cr
@@ -13,6 +13,7 @@ describe "Initializing a new web project" do
       puts "Web app: Running integration spec. This might take awhile...".colorize(:yellow)
       should_run_successfully "lucky init test-project"
       FileUtils.cp("#{src_folder_in_docker}/spec/support/cat.gif", "#{temp_folder}/test-project/public/assets/images/")
+      File.delete("test-project/.env")
       compile_and_run_specs_on_test_project
       File.read("test-project/.travis.yml").should contain "postgresql"
       File.read("test-project/public/mix-manifest.json").should contain "images/cat.gif"
--- a/src/web_app_skeleton/script/setup.ecr
+++ b/src/web_app_skeleton/script/setup.ecr
@@ -34,11 +34,11 @@ if [ -z "$CI" ]; then
 fi
 printf "✔ Done\n" | indent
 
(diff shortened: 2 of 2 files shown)


src/web_app_skeleton/script/setup.ecr
```diff
@@ -34,6 +34,12 @@ if [ -z "$CI" ]; then
 fi
 printf "✔ Done\n" | indent
 
+printf "\n▸ Checking for a .env\n"
```
paulcsmith: I think instead of checking we could print something in the `if` like "Created a .env file"`. That way if you already have one it doesn't say anything. 

But just a suggestion. I'm ok with it as-is if you feel that's better!

spec/integration/init_web_spec.cr
```diff
@@ -13,6 +13,7 @@ describe "Initializing a new web project" do
       puts "Web app: Running integration spec. This might take awhile...".colorize(:yellow)
       should_run_successfully "lucky init test-project"
       FileUtils.cp("#{src_folder_in_docker}/spec/support/cat.gif", "#{temp_folder}/test-project/public/assets/images/")
+      File.delete("test-project/.env")
       compile_and_run_specs_on_test_project
       File.read("test-project/.travis.yml").should contain "postgresql"
       File.read("test-project/public/mix-manifest.json").should contain "images/cat.gif"
```
paulcsmith: Could  you add another `File.exists?` expectation for the env file like the spec below? That way we can be sure it gets created

jwoertink: It already gets checked on line 21. Deleted on 16, created on 17, then checked for existence on 21.

paulcsmith: Another facepalm moment for me :) LGTM

jwoertink: my github says all checks have failed... So not sure what's up with that 🤔

paulcsmith:
Just the Heroku specs. Not sure why it doesn't work for you. For some reason it is not populating the Heroku API keys

PR #369, opened 2019-07-16
Delete duplicate base_flow.cr

matthewmcgarvey:
I believe that the intended `base_flow.cr` is found [here](https://github.com/luckyframework/lucky_cli/blob/master/src/browser_app_skeleton/spec/support/flows/base_flow.cr) and this one can be removed


paulcsmith:
The failed build is unrelated. Gonna go a head and merge this.

Thank you for catching @matthewmcgarvey 🎉

PR #372, opened 2019-07-22
Only add TestBackdoor in test env

paulcsmith:
This way you won't see "test_backdoor" for every action in the logs. That was something that probably makes people nervous


paulcsmith:
Looks like we're not actually signing a user in so I'll have to look into  that

paulcsmith:
Closing in favor of https://github.com/luckyframework/lucky_cli/pull/457

PR #373, opened 2019-07-23
Check settings before attempting migration

paulcsmith:


paulcsmith:
You are right! Thanks @bdtomlin for testing and pointing out CI is just stuck :P

PR #374, opened 2019-07-23
Raise if missing settings as soon as app is loaded

paulcsmith:


stephendolan:
@paulcsmith Is there a way to bump your branch so that the workflows run?

If specs pass, is this ready to merge?

paulcsmith:
I don't think specs pass. I tried this but it didn't work but can't remember why. I won't have time to look at this for awhile so we can close this and open an issue or someone else can take it over in a new PR

PR #376, opened 2019-07-23
adding to setup script checking for postgres

jwoertink:
Closes: https://github.com/luckyframework/lucky/issues/845

This PR adds a check that postgres is both installed, and running prior to running a db.create task. When postgres is installed, but not booted, you'll get a strange error that isn't quite clear what the issue is. By detecting this for the developer, we can save them some time of searching around.

On a side note, I have no clue how to test that postgres isn't running, but I did add this to a new generated lucky app, and killed my local running postgres 

Screen Shot 2019-07-23 at 4 34 10 PM

### DIFF
--- a/src/web_app_skeleton/script/setup.ecr
+++ b/src/web_app_skeleton/script/setup.ecr
@@ -14,10 +14,10 @@ indent() {
 check_postgres() {
   if ! command -v createdb > /dev/null; then
     printf 'Please install the postgres CLI tools, then try again.\n'
-    printf 'See https://www.postgresql.org/docs/current/tutorial-install.html for install instructions.\n'
     if [[ "$OSTYPE" == "darwin"* ]]; then
-      printf "\nIf you're using Postgres.app, see https://postgresapp.com/documentation/cli-tools.html.\n"
+      printf "If you're using Postgres.app, see https://postgresapp.com/documentation/cli-tools.html.\n"
     fi
+    printf 'See https://www.postgresql.org/docs/current/tutorial-install.html for install instructions.\n'
     exit 1
   fi
 
@@ -59,6 +59,7 @@ fi
 
 printf "\n▸ Checking that postgres is installed and running\n"
 check_postgres | indent
+printf "✔ Done\n" | indent
 
 printf "\n▸ Setting up the database\n"
 lucky db.create | indent
--- a/src/web_app_skeleton/script/setup.ecr
+++ b/src/web_app_skeleton/script/setup.ecr
@@ -56,6 +56,9 @@ printf "\n▸ Checking that postgres is installed\n"
 check_postgres | indent
 printf "✔ Done\n" | indent
 
+printf "\n▸ Verifying postgres connection\n"
+lucky db.verify_connection | indent
+
 printf "\n▸ Setting up the database\n"
 lucky db.create | indent
 
--- a/src/web_app_skeleton/script/setup.ecr
+++ b/src/web_app_skeleton/script/setup.ecr
@@ -11,10 +11,13 @@ indent() {
 }
 
 # Ensure postgres is installed, and booted
-ensure_postgres() {
+check_postgres() {
   if ! command -v createdb > /dev/null; then
-    printf 'Please install postgres before continuing.\n'
+    printf 'Please install the postgres CLI tools, then try again.\n'
     printf 'See https://www.postgresql.org/docs/current/tutorial-install.html for install instructions.\n'
+    if [[ "$OSTYPE" == "darwin"* ]]; then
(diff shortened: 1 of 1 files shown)


src/web_app_skeleton/script/setup.ecr
```diff
@@ -10,6 +10,20 @@ indent() {
   done
 }
 
+# Ensure postgres is installed, and booted
+ensure_postgres() {
+  if ! command -v createdb > /dev/null; then
```
paulcsmith: This is a bit tricky because this actually is checking for the CLI tools. You may have Postgres installing, but not the CLI tools. So we should probably be more specific and say Postgres CLI tools are not installed and say if you're using Postges.app to https://postgresapp.com/documentation/cli-tools.html

And then a separate check for `psql`  or  something that  will check if Postgres itself is installed. That should probably go before the `created` check. Does that make sense?

src/web_app_skeleton/script/setup.ecr
```diff
@@ -10,6 +10,20 @@ indent() {
   done
 }
 
+# Ensure postgres is installed, and booted
+ensure_postgres() {
```
paulcsmith: Maybe `check_postgres`?

src/web_app_skeleton/script/setup.ecr
jwoertink: is `psql` a part of the cli tools? I know there's also a `postgres` command for the server. But maybe you don't have that one if you only installed the pg client?

jwoertink: Yup https://www.postgresql.org/docs/11/reference-client.html Looks like those fall under pg client. I guess you don't technically have to have the postgres server installed locally because you could be connecting to a remote DB. And all of the client apps allow you to run them locally and still connect remotely.

I'll update the wording, but I think I'll skip the extra check.

jwoertink:
Here's an updated screenshot of what this would look like. I added a check if you're on a mac, there's a potential that you could be running the Postgres.app, and where you can find some info

Screen Shot 2019-07-23 at 5 01 39 PM

src/web_app_skeleton/script/setup.ecr
```diff
@@ -10,6 +10,23 @@ indent() {
   done
 }
 
+# Ensure postgres is installed, and booted
+check_postgres() {
+  if ! command -v createdb > /dev/null; then
+    printf 'Please install the postgres CLI tools, then try again.\n'
+    printf 'See https://www.postgresql.org/docs/current/tutorial-install.html for install instructions.\n'
+    if [[ "$OSTYPE" == "darwin"* ]]; then
```
paulcsmith: Let's put this above the other  once since if it matches it is more likely to be the one people want. A ton of people use Postgres.app and if they don't then they can check out the other instructions.

src/web_app_skeleton/script/setup.ecr
paulcsmith: Sounds good!

src/web_app_skeleton/script/setup.ecr
```diff
@@ -40,6 +57,9 @@ if [ ! -f ".env" ]; then
   printf "✔ Done\n" | indent
 fi
 
+printf "\n▸ Checking that postgres is installed and running\n"
```
paulcsmith: Maybe do what we did on line 57 and print  done after the check so it is clear it succeeded?

paulcsmith:
Hmm seems like tests  in CI are failing because it says "Postgres is not running yet". Seems like the check is not working for some reason :S

Screen Shot 2019-07-24 at 11 15 59 AM

jwoertink:
lovely... This is gonna be tricky to check....

paulcsmith:
Yeah :S

BTW I added another thing we can check in script/setup! https://github.com/luckyframework/lucky_cli/issues/377

jwoertink:
Ok, here's the deal. It turns out that you don't have to have postgres running for all this to work. Since you're able to connect to a remote running instance, you only really have to have the CLI tools installed to do that. 

The new plan will be in 3 stages.
* Stage 1: check that postgres is installed through this setup script
* Stage 2: Add a new task to Avram that will check for a connection using whatever you have setup
* Stage 3: Come back to this script and add the new task usage.

This last commit removes the check that postgres is running, but only checks that `createdb` is installed. I can't imagine that you'd ever have only `createdb` and not `dropdb` or `psql` installed. If that super weird edge case ever comes up, then we can add additional checks I guess, but for now I don't think that's necessary.

paulcsmith:
Thanks for the summary of next steps. Sounds like a good plan to me!

PR #386, opened 2019-08-12
Use VerbNoun convention for operations

### LINKED ISSUE #385: Rename operations to use VerbNoun
FYI @jwoertink @edwardloveall

paulcsmith:
Conventions make the world a better place. Let's use them :D

Closes #385

~I left `ResetPassword` and `RequestPasswordReset`. I think using just a Verb in some cases. Thoughts?~ I was trying to figure out how to use `User` but forgot we still had a noun there 🤦‍♂️


jwoertink:
Wouldn't `Password` be the noun? So `ResetPassword` still fits this convention. And in the second one a `PasswordReset` would be the noun, so that one also fits this convention 😉

paulcsmith:
I've made a huge mistake

jwoertink:
Good Gob!

PR #389, opened 2019-08-15
Add accepted formats

paulcsmith:


paulcsmith:
I will need to look at why CI thinks it is not formatted. Locally the check passes fine. Going to go ahead and merge for now

PR #391, opened 2019-08-22
Verify connection after db create

paulcsmith:
The problem was that we'd verify the connection before the db was
created.

This meant that it would always raise until the db.create ran.

### DIFF
--- a/src/web_app_skeleton/script/setup.ecr
+++ b/src/web_app_skeleton/script/setup.ecr
@@ -56,11 +56,11 @@ printf "\n▸ Checking that postgres is installed\n"
 check_postgres | indent
 printf "✔ Done\n" | indent
 
+printf "\n▸ Creating the database\n"
+lucky db.create | indent
+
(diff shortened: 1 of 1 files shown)


src/web_app_skeleton/script/setup.ecr
```diff
@@ -56,12 +56,12 @@ printf "\n▸ Checking that postgres is installed\n"
 check_postgres | indent
 printf "✔ Done\n" | indent
 
+printf "\n▸ Creating the database\n"
+lucky db.create | indent
```
paulcsmith: This  should work since we check that Postgres `createdb` is installed. Still possible that Postgres is not running and `db.create` fails, but that is better than  verify_connection failing every time on first run of script/setup

paulcsmith:
It's a Lucky CLI issue I think. When a task raises it does not exit with a non-zero status so the bash script continues. I'll try to fix that later :D

PR #406, opened 2019-09-16
Refactor lucky init

### LINKED ISSUE #405: Check for a valid project name before asking other questions
Currently when generating a new project you can enter whatever you want as the name, it then asks you a couple questions and then, if the name isn't valid, fails to work. It would be nice if it validated the name at the beginning.

jwoertink:
Fixes #405 

This PR changes up a few things.
* Remove argument passing to `lucky init` command. Our guides only mention using the wizard anyway, and use of `lucky init project` can lead to some confusion. 
* Add a welcome banner, because it's friendly, but also lets you know which version of a lucky app you're about to generate
* Delays loading of the questions so that way project name validation happens directly after it's been asked, and not all the way at the end.
* Adds a few helpful comments for project name to help avoid having to start over.
* Moves tasks.cr file generation from `lucky init tasks.cr` to `lucky gen.tasks_file`.
* Adds custom init command to setup a project without going through the wizard with `lucky init.custom my_project`

out


src/init.cr
```diff
@@ -6,38 +6,13 @@ class LuckyCli::Init
   end
 
   def run
-    project_name = name
```
paulcsmith: The  `name` private method should also be removed and maybe it should error if any args are provided?

src/init.cr
```diff
@@ -6,38 +6,13 @@ class LuckyCli::Init
...
-      LuckyCli::Generators::Web.run(
-        project_name: project_name,
-        options: options
-      )
     else
       LuckyCli::Wizard::Web.run
     end
   end
 
-  private def options : Options
```
paulcsmith: We should still allow using these options but under a different command name. There are times using the CLI is helpful for stuff like this and a wizard would make things pretty hard. Maybe `lucky init.custom` or something. I don't really have a good name for it :S

src/lucky_cli/wizard/web.cr
```diff
@@ -9,28 +9,35 @@ class LuckyCli::Wizard::Web
...
       project_name: ask_for_project_name,
       options: options
     )
   end
 
-  private def ask_for_project_name
-    TextQuestion.ask("Project name?").tap do
-      puts "\n"
-    end
+  private def ask_for_project_name : Proc(String)
```
paulcsmith: I think rather than passing procs around, it may be simpler to extend `TestQuestion` with an optional `validator` proc that can validate the result. It'd go here where we validate that it is present: https://github.com/luckyframework/lucky_cli/blob/db7a0a478fc7cc422b42a011d2e6710f83196db4/src/lucky_cli/wizard/text_question.cr#L14-L19

Maybe the Proc would be `Proc(String, Error?)` where `Error` is a super  simple class that holds an error message. If it returns one then it prints the error message and asks again.

This approach would also mean that  you could use `TextQuestion` with other kinds of validations or later open this up to your own tasks that need validations

src/init.cr
jwoertink: From the look of it, there can still be an argument for the task stuff. We could move that out of `init` and in to something else. So right now `lucky init tasks.cr` generates a new task file. We could do something like `lucky generate.task` or something?

paulcsmith: I like the idea of a different task. Maybe `lucky gen.tasks_file` to is it clear it is not a task

src/init.cr
jwoertink: It's called `init.custom` for now. We can always rename later if we have a better name 😅

src/lucky_cli/wizard/web.cr
jwoertink: I'm having a little trouble seeing this. Right now there's 2 types of validations that happen on `TextQuestion`. 1 being `answer.empty?` right there, and the other being done in this `Validators::ProjectName` module. Each have their own validation rule, and error message. It sounds like we'd need a list of validation rules, or a collection of validators and not just a single one. But is this all overkill to abstract all this TextQuestion stuff? We only use it once, and is this something you see people being able to hook in to for use in their tasks?

If this is to be used as an additional tool for tasks, then we should probably re-structure all this anyway to remove it from the `Wizard` namespace. If not, then I'm thinking just throw it all in the class as if it's only used in this one spot any way.... What are your thoughts?

paulcsmith: I think I get what you’re saying and I agree. Probably doesn’t make sense to abstract validations if it’s only used for the project name. How about renaming it to PorjectNameQuestion and moving the validation there. Like we do with “empty?”

I think that’s what you were saying right? Or something along those lines?

jwoertink: Yeah, exactly. Ok, so if it's just going to be used for this one instance, then we can just move things all in to this class, and clean up the logic a bit.

paulcsmith: Yeah let’s do that. We can abstract later if we want but for now let’s do simple. Good call

jwoertink:
Locally I'm getting a few failing specs, but not sure why yet...

```
In lib/lucky/src/lucky/error_action.cr:36:16

 36 | abstract def render(error : Exception) : Lucky::Response
                   ^-----
Error: abstract `def Lucky::ErrorAction#render(error : Exception)` must be implemented by Errors::Show
```

```
  1) Initializing a new web project creates a full web app successfully
     Failure/Error: result.exit_status.should be_successful

       Expected: 0
            got: 256

     # spec/support/should_run_successfully.cr:11
```

These may be related to the fact that the specs tell docker to install the cli from master, but I merged in master.... They may also be unrelated to this PR. I'll have to dig around more.

paulcsmith:
Make sure you've rebased. I recently made changes that should fix this

jwoertink:
I'm all merged in though. That's the part kind of confusing me. My guess is it has to be some weird docker cache. I can try wiping that all out, and starting again, but not sure about travis.

paulcsmith:
Could also be that I made some changes recently to Lucky. I'm not sure. I'm gonna get #409 merged in and see if that does it

paulcsmith:
Rebased and a few small refactors here https://github.com/luckyframework/lucky_cli/pull/420

PR #407, opened 2019-09-16
Rename 'render' to 'html'. 

paulcsmith:


paulcsmith:
Merging even though CI  is red because it  is  red for a  different reason than this PR

PR #414, opened 2019-09-24
[DO NOT MERGE] Jaw/v0.18.0

jwoertink:
This is basically #413 but branched from c45e1860751bba25a16120402f93e7537c0be5b5


src/web_app_skeleton/.crystal-version
```diff
@@ -1 +1 @@
-0.30.0
+0.30.1
```
paulcsmith: I  think this is supposed toe 0.31.0

src/lucky_cli/version.cr
```diff
@@ -1,3 +1,3 @@
 module LuckyCli
-  VERSION = "0.17.0-rc1"
+  VERSION = "0.17.0"
```
paulcsmith: Should this  be 0.18.0?

src/web_app_skeleton/.crystal-version
jwoertink: lol yup. 😅

src/lucky_cli/version.cr
jwoertink: ugh 😫 yeah. I knew I'd miss stuff.

paulcsmith:
I  think we should also sneak this into this release since it fixes the issue where it verifies the connection before the db is created https://github.com/luckyframework/lucky_cli/commit/aa2cb7a4946e59cb07b83fadd773d0b2cfaa7003#diff-154806d6e0faa0b1aa1e518f3bbd3647

paulcsmith:
We've decided to make 0.18 The Big One™  so we'll close this for now

PR #420, opened 2019-09-27
Remove option to use init with args

### LINKED ISSUE #405: Check for a valid project name before asking other questions

paulcsmith:
This is #406 with some tweaks

And introduce a new task `custom` to use the CLI and not the
wizard.

This also moves validation of project name to be immediate.

Closes #405 #406 

Screen Shot 2019-09-26 at 8 43 21 PM


spec/lucky_cli/project_name_spec.cr
```diff
@@ -0,0 +1,39 @@
...
+    it "doesn't allow uppercase characters" do
+      validate(title).should be_false
+    end
+
+    it "corrects uppercase characters" do
+      expected = "from_the_earth_to_the_moon"
+      sanitize(title).should eq expected
+    end
+  end
+end
```
paulcsmith: This is mostly unchanged from #406. Just  uses `ProjectName` instead of `ProjectNameQuestion`

src/custom_init.cr
```diff
@@ -0,0 +1,32 @@
+class LuckyCli::CustomInit < LuckyCli::Init
+  def run
+    project_name = ProjectName.new(ARGV[1]? || "")
```
paulcsmith: Sanitization, validation, and validation messages are now in the `ProjectName` object

src/custom_init.cr
```diff
@@ -0,0 +1,32 @@
...
+      parser.on("-h", "--help", "This help message") {
+        puts parser
+        exit(0)
+      }
+    end
+
+    LuckyCli::Generators::Web.run(
+      project_name: project_name.to_s,
+      api_only: api_only,
+      generate_auth: authentication
```
paulcsmith: Decided to remove the `Options` object. It was more annoying than helpful. Not sure why I added it in the first place :)

src/lucky_cli/project_name.cr
```diff
@@ -0,0 +1,33 @@
...
+
+      How about: '#{sanitized_name}'?
+      TEXT
+    end
+  end
+
+  def sanitized_name
+    name.downcase.gsub(/[^a-z0-9_-]/, "_").strip('_')
+  end
+end
```
paulcsmith: Validations and sanitize are now in one spot and can be used by the wizard or through the CLI

paulcsmith:
I opted to leave out the validation prompt because I felt it was a bit overwhelming to have that big list of requirements. It'll still show you what the requirements are if you don't quite get it. Thoughts?

I love the addition of the welcome message. Good idea @jwoertink

paulcsmith:
No the 0.31.0 changes are in another PR and not merged yet. I’ll double check Travis though to make sure

PR #425, opened 2019-10-07
Update JS deps

### LINKED ISSUE #394: Rails UJs update?
I just came across this https://yarnpkg.com/en/package/@rails/ujs which shows a higher version number than https://yarnpkg.com/en/package/rails-ujs. Not sure if they are moving over, or what's up with that. Maybe we use this new one? Maybe we just roll our own LuckyUJS? No clue, but I wanted to just put this out there.

vlazar:
Fixes #394

In addition I've also updated Turbolinks. Rails UJS and Turbolinks are both from Rails and I think it makes sense to update both to latest versions.

I also wonder if other JS dependencies should be updated too?

Actually maybe I miss something, but is there a reason why laravel-mix's dependencies are specified explicitly and some are locked to specific version? Just a guess, but looks like everything below `"laravel-mix": "^4.0.0",` shouldn't be required.

```js
  "devDependencies": {
    "browser-sync": "^2.18.13",
    "laravel-mix": "^4.0.0",
    "resolve-url-loader": "2.3.1",
    "sass": "1.17.1",
    "sass-loader": "7.*",
    "vue-template-compiler": "^2.5.22"
  }
```

### DIFF
--- a/src/browser_app_skeleton/package.json
+++ b/src/browser_app_skeleton/package.json
@@ -4,7 +4,7 @@
   "dependencies": {
     "@rails/ujs": "^6.0.0",
     "normalize-scss": "^7.0.1",
-    "turbolinks": "^5.0.3"
+    "turbolinks": "^5.2.0"
   },
   "scripts": {
     "heroku-postbuild": "yarn prod",
--- a/src/browser_app_skeleton/package.json
+++ b/src/browser_app_skeleton/package.json
@@ -2,8 +2,8 @@
   "license": "UNLICENSED",
   "private": true,
   "dependencies": {
+    "@rails/ujs": "^6.0.0",
     "normalize-scss": "^7.0.1",
-    "rails-ujs": "^5.1.4",
     "turbolinks": "^5.0.3"
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


jwoertink:
Thanks for help on all these PRs!

My personal opinion is that all libs should always be locked to specific versions. Lucky has already done this with 3rd party shards we don't control on the crystal side.

This helps to save in cases where you have a new team member, or maybe you need to re-setup your app, and you end up getting new versions of things that don't quite work with your setup. Then you're either forced to update your app, or figure out which versions you can rollback to. 

Let's see what @paulcsmith thought is on this before deciding which way to go.

vlazar:
Yeah, dependencies are hard. JS dependencies are hell. Maybe locking JS dependencies in generated `package.json` will add extra level of safety for Lucky users in case some minor update of `laravel-mix` mathching `^4.0.0` version restriction breaks something.

It's strange though that `laravel-mix` is not locked to specific version, but some of it's dependencies are.

jwoertink:
I think this is fine for now. We can always update to lock in later. Thanks for adding this in!

PR #427, opened 2019-10-11
Stop logging favicon requests in development

vlazar:
I've noticed `GET /favicon.ico` requests sometimes shown in development mode.

Since requests to all other assets are configured to be skipped, I think it makes sense to skip `/favicon.ico` as well.

I'm not sure however if other extensions for favicon should be specified too. As stated in https://en.wikipedia.org/wiki/Favicon other image formats like `.png`, `.gif` and so on can be used. Even path can be different.

Since the default template does not include any meta tags for favicon I think having only the default `favicon.ico` in regex might be sufficient.

**WDYT?**

As a side note **I've found out about this behavior when I was literally pulling my hair out** trying to figure out why `Lucky.logger.info` was printing info multiple times 😕 (sometimes 3, sometimes 5 🤔) when used from HTTP handler. 

Since all requests to assets are not logged and `favicon.ico` requests are rare, it took me awhile to realize that browser sometimes sends requests to just HTML and sometimes for several assets too and every sent requests result in ` more log entry.

If logging for assets would be on in development I would have noticed this quicker for sure. However since I was only seeing a single `GET /sign_in` and then all messages under it it wasn't obvious.

<details>
  <summary>See example logs</summary>

12:59:07 PM web.1   |  GET /sign_in
12:59:07 PM web.1   |   ▸ Ran FooHandler#call 1
12:59:07 PM web.1   |   ▸ Handled by SignIns::New
12:59:07 PM web.1   |   ▸ Ran verify_accepted_format
12:59:07 PM web.1   |   ▸ Ran protect_from_forgery
12:59:07 PM web.1   |   ▸ Ran test_backdoor
12:59:07 PM web.1   |   ▸ Ran redirect_signed_in_users
12:59:07 PM web.1   |   ▸ Rendered SignIns::NewPage
12:59:07 PM web.1   |   ▸ Sent 200 (1.47ms)
12:59:07 PM web.1   |   ▸ Ran FooHandler#call 2
12:59:07 PM web.1   |   ▸ Ran FooHandler#call 3
12:59:07 PM web.1   |   ▸ Ran FooHandler#call 4
</details>

**So I guess my question is - could/should something be done for my use case (logging from HTTP handlers) to avoid confusion for someone else?**

As a second side note I found this cool tool which allows inspect requests between browser and server and much more. You probably already knew, but I thought I would mention it anyways [HTTP Toolkit](https://httptoolkit.tech/)

### DIFF
--- a/src/web_app_skeleton/config/logger.cr
+++ b/src/web_app_skeleton/config/logger.cr
@@ -38,7 +38,7 @@ Lucky::LogHandler.configure do |settings|
   if Lucky::Env.development?
     settings.skip_if = ->(context : HTTP::Server::Context) {
       context.request.method.downcase == "get" &&
-      context.request.resource.starts_with?(/\/css\/|\/js\/|\/assets\//)
+      context.request.resource.starts_with?(/\/css\/|\/js\/|\/assets\/|\/favicon\.ico/)
     }
   end
 end


vlazar:
Icons mentioned #303 are probably also need to be taken care of in logger.

edwardloveall:
I don't want to bog this down, but I do sometimes need to debug the favicon and it's possible that knowing that the request was being made would be helpful. What do you think about making this configurable?

vlazar:
@edwardloveall This change is in a config for generated web app (`src/web_app_skeleton/config/logger.cr`) so it is configurable.

edwardloveall:
Oh right, great point. Sorry, I loose track of which repo I'm looking at.

src/web_app_skeleton/config/logger.cr
```diff
@@ -38,7 +38,7 @@ Lucky::LogHandler.configure do |settings|
   if Lucky::Env.development?
     settings.skip_if = ->(context : HTTP::Server::Context) {
       context.request.method.downcase == "get" &&
-      context.request.resource.starts_with?(/\/css\/|\/js\/|\/assets\//)
+      context.request.resource.starts_with?(/\/css\/|\/js\/|\/assets\/|\/favicon\.ico\//)
```
jwoertink: This wouldn't match `/favicon.ico` https://play.crystal-lang.org/#/r/7u9a That last slash should go away.

vlazar: 🤦‍♂ Now that's embarrassing. Thanks for catching this! I've checked that it works in logs, but later I did some other experiments and it ended up like this.

jwoertink: no worries!

vlazar:
@jwoertink BTW, do you have any comments on other logging concerns I've mentioned in PR description?

jwoertink:
Oh, so you're saying that even though the css wasn't logged, the handler stack calls to that css still are?

vlazar:
Yeah, because I was logging from handler classes and it was super confusing to me especially with nested logging. I think that's why I've spend more time on this than I could have otherwise.

jwoertink:
Ok, that's a separate issue. Let's add that as a new issue, and then maybe reference https://github.com/luckyframework/lucky/issues/950 as a side note.

PR #428, opened 2019-10-16
Moving the error handler above session in the stack. 

jwoertink:
This lets the cookie overflow error to render the exception page. Maybe it can move all the way up? Not sure if that matters or not, but moving to this position does fix the issue from https://github.com/luckyframework/lucky/issues/949

### DIFF
--- a/src/web_app_skeleton/src/app_server.cr.ecr
+++ b/src/web_app_skeleton/src/app_server.cr.ecr
@@ -1,10 +1,10 @@
 class AppServer < Lucky::BaseAppServer
   def middleware : Array(HTTP::Handler)
     [
+      Lucky::ErrorHandler.new(action: Errors::Show),
       Lucky::ForceSSLHandler.new,
       Lucky::HttpMethodOverrideHandler.new,
       Lucky::LogHandler.new,
-      Lucky::ErrorHandler.new(action: Errors::Show),
       <%- if browser? -%>
       Lucky::SessionHandler.new,
       Lucky::FlashHandler.new,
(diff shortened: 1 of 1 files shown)


jwoertink:
Ok, this one is a little tricky. I moved the error handler to the top of the stack which fixed the issue of throwing the proper page when errors occur in other parts of the stack; however, this caused a new issue for me. I set a flash in my error action, and the flash doesn't show. This is because flash is set later in the stack, so it's never actually stored when an error occurs. 

Easy fix is put flash first, then error second. Though, that makes me wonder if there's other things that might come up. What should the actual stack look like? 

```crystal
# current proposal
Lucky::FlashHandler.new,
Lucky::ErrorHandler.new(action: Errors::Show),
Lucky::ForceSSLHandler.new,
Lucky::HttpMethodOverrideHandler.new,
Lucky::LogHandler.new,
Lucky::SessionHandler.new,
Lucky::RouteHandler.new,
Lucky::StaticFileHandler.new("./public", false),
Lucky::RouteNotFoundHandler.new,
```

paulcsmith:
Hmm I think we should see if we can modify FlashHandler and SessionHandler so that it works with the ErrorHandler. I'm not totally sure I get why they don't. Maybe they need an `ensure` in there so it tries to write the cookies on error but still bubble up the error to ErrorHandler?

jwoertink:
This might not be an issue anymore. It would seem to make sense that the ErrorHandler would come first, but I just tried in a fresh app, and everything works as expected as is. 🤷‍♂ I can throw a cookie overflow, and it renders the error page just fine. Not sure what was up with my project. 

One thing I did notice was by moving the ErrorHandler to the top, that causes the log to show the exception twice. I think I'm gonna close this since making these changes in a fresh app doesn't seem to give me anything useful.

paulcsmith:
Sounds good! Thanks for testing it out

PR #445, opened 2019-12-27
Add config folder to format check and fix files incorrectly formatted

matthewmcgarvey:
Just a minor nitpick but I saw that there was already a check for formatting so it seems to be the intention to have the generated code be correctly formatted.

I am not happy with the formatting on `cookies.cr.ecr`.  I will see if I can replicate it in a smaller codebase and open an issue on crystal for it.

I moved the check up since it doesn't require running `script/setup` and it will fail fast which is a whole lot better than waiting several minutes to find out the formatting is incorrect.

### DIFF
--- a/spec/integration/init_web_spec.cr
+++ b/spec/integration/init_web_spec.cr
@@ -102,8 +102,8 @@ end
 
 private def compile_and_run_specs_on_test_project
   FileUtils.cd "test-project" do
+    should_run_successfully "crystal tool format --check spec src config"
     should_run_successfully "script/setup"
-    should_run_successfully "crystal tool format --check spec src"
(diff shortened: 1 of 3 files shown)


matthewmcgarvey:
Crystal issue: https://github.com/crystal-lang/crystal/issues/8630

paulcsmith:
Hi @matthewmcgarvey! I thought I had responded, but I think I forgot to actually post it!

Thanks for opening this and the issue on Crystal. I agree the formatting is a bit wonky on cookies.cr. So I'll hold off on merging until the issue is fixed in Crystal. I subscribed to the issue so I should get notified when it is closed!

jwoertink:
The next release of crystal will have that formatting fixed.

PR #457, opened 2020-03-11
Only add TestBackdoor in test env

paulcsmith:
This way you won't see "test_backdoor" for every action in the logs.
That was something that probably makes people nervous

### DIFF
--- a/src/browser_authentication_app_skeleton/src/actions/mixins/auth/test_backdoor.cr
+++ b/src/browser_authentication_app_skeleton/src/actions/mixins/auth/test_backdoor.cr
@@ -1,5 +1,11 @@
+# When testing you can skip normal sign in by using `visit` with the `as` param
+#
+#     user = UserBox.create
+#     flow.visit Me::Show, as: user
+#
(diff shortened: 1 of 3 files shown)


src/browser_authentication_app_skeleton/src/actions/mixins/auth/test_backdoor.cr
```diff
@@ -1,13 +1,21 @@
...
+#
+#     user = UserBox.create
+#     flow.visit Me::Show, as: user
+#
+# The module only works in the 'test' environment.
 module Auth::TestBackdoor
-  macro included
-    before test_backdoor
-  end
+  {% if env("LUCKY_ENV") == "test" %}
```
paulcsmith: Does not work because LUCKY_ENV is set at runtime in spec_helper, so is not available at this point :(

paulcsmith:
We no longer need this. The main reason for it was so it didn't log, but now we only log if halted: https://github.com/luckyframework/lucky/pull/1062

PR #461, opened 2020-03-13
Change SignIn -> LogIn

### LINKED ISSUE #456: Rename SignUp to Register or CreateAccount
SignUp and SignIn are very similar and it takes a moment to figure out what they are doing, especially for non-native English speakers. We should use a different term for "creating an account". I think a common one is "Register" or just "Create Account"

* `SignUps::New/Create` -> `Registrations::New/Create`
* `SignUpUser` -> `RegisterUser`

paulcsmith:
Closes #456 

Looking for feedback on naming. This is broken because we'd also need to change Authentic#sign_in to Authentic#log_in


paulcsmith:
Also @edwardloveall I'm not sure when to use Log in versus Login. Since you're good at English maybe you could help me out 🤣

src/api_authentication_app_skeleton/spec/requests/api/log_ins/create_spec.cr
```diff
@@ -1,11 +1,11 @@
 require "../../../spec_helper"
 
-describe Api::SignIns::Create do
+describe Api::LogIns::Create do
```
paulcsmith: I guess this should be `Logins` since it is a noun? https://grammarist.com/spelling/log-in-login/ but I think operations should still be a verb so `LogInUser` is correct...I think..

jwoertink: hmm.. yeah, I think `Logins::Create` would be more appropriate here.

src/api_authentication_app_skeleton/src/actions/mixins/api/auth/skip_require_auth_token.cr
```diff
@@ -3,7 +3,7 @@ module Api::Auth::SkipRequireAuthToken
     skip require_auth_token
   end
 
-  # Since sign in is not required, current_user might be nil
+  # Since log in is not required, current_user might be nil
```
jwoertink: Maybe `# Since authentication is not required, `? or `Since login is not required`

src/browser_authentication_app_skeleton/src/actions/log_ins/delete.cr
```diff
@@ -0,0 +1,7 @@
+class LogIns::Delete < BrowserAction
+  delete "/sign_out" do
```
jwoertink: Should this be `/logout` or `/log_out`?

src/browser_authentication_app_skeleton/src/actions/log_ins/delete.cr
```diff
@@ -0,0 +1,7 @@
...
+    sign_out
```
jwoertink: Same here. It seems a little out of place to have `Logins::Delete` call `sign_out`.

edwardloveall:
I do have opinions about this!

I think trying to make Login or Signup a noun is what's difficult. How about `Authentications`/`Registrations`? Then we can keep the restful wording of `Create` and `Destroy`.

I'm not super into the LogX stuff. I think @jwoertink pointed out a common confusion about if it's one or multiple words. [Thoughts](https://english.stackexchange.com/questions/5302/log-in-to-or-log-into-or-login-to) that I found on that are... confusing at best.

paulcsmith:
English...why u so hard to understand!! Maybe `Sessions` and `Registrations`? I kind of like how signins/logins matches the actual language in the UI, but maybe not worth it!

edwardloveall:
`Sessions` works too!

I hear you on code not matching the UI. It's difficult because we don't talk about the noun ever. I.e. you `sign in` (verb) and now you are `signed in` (adjective) but no one uses the noun version, which isn't `you have a sign in` it's `you have a session`.

edwardloveall:
Now you got me thinking about this:

We like when [operations are VerbNoun](https://github.com/search?q=verbnoun+org%3Aluckyframework&type=Issues)s and it seems like actions are best as PluralNoun::Verb like `Users::Create`. Maybe we could encode this somewhere 🤔

paulcsmith:
Ok based in this here is my new proposal:

* `SignIns` -> `Sessions`
* `SignUps` -> Stays the same?

Language in the UI:

* `Sign In` -> `Log In`
* `Sign Up` -> Stays the same

So my main thought is `Sign Up` still fits pretty well and is used a lot. But open to thoughts on this. I could also see us changing to `Registrations` in the code and `Sign Up` in the UI...but I'm not sure

paulcsmith:
Oh and yes I think you're right about that convention @edwardloveall, but I think we can hold off because of that discussion on potential changes to the file structure that I think would be a lot more consistent `Noun::Verb` for operations and actions. And just `Noun` for pages and others (like it already is) https://github.com/luckyframework/lucky/issues/1030#issuecomment-597721197

edwardloveall:
I still think `SignUp` is not really a noun, (I remembered the word for this btw, I think it's a [gerund](https://en.wikipedia.org/wiki/Gerund)) but it reads fine enough that I think everyone will get it.

paulcsmith:
@edwardloveall Yeah it is kinda weird. And it also might be weird to have `SignUps` if the UI text is changed to `Create an account` or something. `Registrations` seems more generic.

Thoughts on just going with that instead of SignUps in code, but keeping `Sign up` in the UI

edwardloveall:
Sounds great to me!

stephendolan:
@paulcsmith Is this something you'd like someone else to grab and close out?

paulcsmith:
@stephendolan I think we can close this since it is so out of date, but I'm not sure what we should name these. I never felt super good about it. So I think we can continue this in a dscussion maybe

PR #464, opened 2020-03-14
Adding ability to set custom directory when using custom.init

### LINKED ISSUE #339: [Feature] Allow to specify output directory for init via command line option
The command `lucky init [name]` creates a new new project in `./[name]`. (Same holds when executing the wizard). 

Perhaps there is a chance to expand the command line arguments the cli tool takes to be able to specify a differently named output directory, which would override `@project_dir = project_name` in the generator?

As a special corner case, I would like to be able to use the current directory as the target.

jwoertink:
Fixes #339

Right now we have a `custom.init` generator that I don't think is used much. It only does the same as the wizard, but all through the command line. Now we can start to allow for some really custom options like setting a custom directory!

This PR adds a new `--dir DIR` flag to set the directory of where to generate your application. All it really does is cd in to that custom directory, and then generate the app as normal from that directory. This may be helpful for people using docker, or automating a build on to a server.

```
lucky custom.init my-project --dir /home/me/projects
```

### DIFF
--- a/spec/integration/deploy_to_heroku_spec.cr
+++ b/spec/integration/deploy_to_heroku_spec.cr
@@ -38,6 +38,7 @@ require "../spec_helper"
   private def generate_heroku_app_name
     app_name_base + "-" + Random::Secure.hex(4)
   end
+
   private def app_name_base
     ENV["HEROKU_APP_NAME"]? || "lucky-integration"
   end
--- a/spec/integration/init_web_spec.cr
+++ b/spec/integration/init_web_spec.cr
@@ -62,7 +62,7 @@ describe "Initializing a new web project" do
   end
 
   it "creates a full app in a different directory" do
-    puts "Web app with custom directory: Running integration spec. This might take awhile...".colorize(:yellow)
+    puts "Web app with custom directory: Running integration spec.".colorize(:yellow)
     FileUtils.mkdir_p "/tmp/home/bob"
     should_run_successfully "crystal run src/lucky.cr -- init.custom test-project --dir /tmp/home/bob"
     FileUtils.cd "/tmp/home/bob/test-project" do
@@ -74,7 +74,7 @@ describe "Initializing a new web project" do
     FileUtils.mkdir "test-project"
     output = IO::Memory.new
     Process.run(
-      "crystal run src/lucky.cr -- init.custom test-project test-project",
+      "crystal run src/lucky.cr -- init.custom test-project",
       output: output,
       shell: true
     )
@@ -97,7 +97,7 @@ describe "Initializing a new web project" do
   it "translates dashes to underscores in the project name" do
     output = IO::Memory.new
     Process.run(
-      "crystal run src/lucky.cr -- init.custom test-project 'test-project'",
+      "crystal run src/lucky.cr -- init.custom test-project",
       env: ENV.to_h,
       output: output,
       shell: true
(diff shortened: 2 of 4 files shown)


spec/integration/deploy_to_heroku_spec.cr
```diff
@@ -38,6 +38,7 @@ require "../spec_helper"
   private def generate_heroku_app_name
     app_name_base + "-" + Random::Secure.hex(4)
   end
+
```
jwoertink: the crystal formatter did this

spec/integration/init_web_spec.cr
```diff
@@ -9,7 +9,7 @@ end
 describe "Initializing a new web project" do
   it "creates a full web app successfully" do
     puts "Web app: Running integration spec. This might take awhile...".colorize(:yellow)
-    should_run_successfully "crystal run src/lucky.cr -- init.custom test-project test-project"
```
jwoertink: no clue why we had this doubled on all these specs lol

src/custom_init.cr
```diff
@@ -13,14 +13,18 @@ class LuckyCli::CustomInit < LuckyCli::Init
   private def generate_project(project_name)
     api_only = false
     authentication = true
+    project_directory = "."
```
jwoertink: Just default it to the current directory you're in like it does now

src/generators/web.cr
```diff
@@ -9,8 +9,11 @@ class LuckyCli::Generators::Web
   def initialize(
     project_name : String,
     @api_only : Bool,
-    @generate_auth : Bool
+    @generate_auth : Bool,
+    project_directory : String = "."
   )
+    Dir.cd(File.expand_path(project_directory))
```
jwoertink: The `expand_path` allows us to pass things like `"."` or `~/Projects` and have it expand properly. We just cd in to the directory we want the project, then run everything as normal.

spec/integration/init_web_spec.cr
paulcsmith: britney-spears-confused

spec/integration/init_web_spec.cr
```diff
@@ -45,22 +45,31 @@ describe "Initializing a new web project" do
...
 
   it "creates a full app without auth" do
     puts "Web app without auth: Running integration spec. This might take awhile...".colorize(:yellow)
-    should_run_successfully "crystal run src/lucky.cr -- init.custom test-project test-project --no-auth"
+    should_run_successfully "crystal run src/lucky.cr -- init.custom test-project --no-auth"
     compile_and_run_specs_on_test_project
   end
 
+  it "creates a full app in a different directory" do
+    puts "Web app with custom directory: Running integration spec. This might take awhile...".colorize(:yellow)
```
paulcsmith: You can probably remove the part about it taking awhile. I think this should be a second or two at most

src/custom_init.cr
```diff
@@ -13,14 +13,18 @@ class LuckyCli::CustomInit < LuckyCli::Init
...
     OptionParser.parse do |parser|
       parser.banner = <<-TEXT
       Usage: lucky custom.init NAME [arguments]
 
-      Example: lucky custom.init my_api --api --no-auth
+      Example: lucky custom.init my_api --api --no-auth --dir ~/Projects/
```
paulcsmith: Maybe add a note that all these arguments are optional and all it requires is the project name

jwoertink: How about 2 examples. 1 that shows with just the name, and a 2nd with the options?

paulcsmith: Yeah that sounds good!

PR #465, opened 2020-03-14
Change development ENV from PORT to DEV_PORT

### LINKED ISSUE #324: PORT confusion
I needed my app to boot on a different port, so I updated my `config/watch.yml` file to port 5020 and ran `crystal src/server.cr`. The app booted on 5020!

Then since my app uses multiple services, I needed to boot it with `lucky dev`. When I ran that, the app booted on port 5000.... This was very confusing since my setting was `settings.port = Lucky::ServerSettings.port`. Taking a look at how [that is defined](https://github.com/luckyframework/lucky/blob/master/src/lucky/server_settings.cr#L13) shows me that it looks at the env var first, but I didn't set any env var... 

After 30min of digging around I come to find out that [overmind](https://github.com/DarthSim/overmind#specifying-the-ports) looks for a `$PORT` var first, then sets it to 5000 if it doesn't see it. This renders the `watch.yml` file almost useless (except for the host).

Since the cli [supports overmind](https://github.com/luckyframework/lucky_cli/blob/master/src/lucky_cli/process_runner.cr#L41) by default, I think we need to come up with a fix for this.

1. Look in to getting rid of the watch.yml file... We would need to figure out how browser sync would work with all this I think, but otherwise not horrible.

2. Find a way to pass the `-p` option down to overmind. This may be difficult just with the separation of where the data comes from.

3. Raise an exception in development if the app boots on a different port than what's in the watch.yml. This would be pretty annoying and I don't really like this as a solution, but hey, it's an option I guess..

jwoertink:
Fixes #324

See also https://github.com/luckyframework/lucky/pull/1051

### DIFF
--- a/src/web_app_skeleton/config/server.cr.ecr
+++ b/src/web_app_skeleton/config/server.cr.ecr
@@ -16,7 +16,7 @@ Lucky::Server.configure do |settings|
   else
     settings.secret_key_base = "<%= secret_key_base %>"
     # Change host/port in config/watch.yml
-    # Alternatively, you can set the PORT env to set the port
+    # Alternatively, you can set the DEV_PORT env to set the port for local development
     settings.host = Lucky::ServerSettings.host
     settings.port = Lucky::ServerSettings.port
   end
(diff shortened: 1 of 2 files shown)


paulcsmith:
Might be worth putting “breaking change” in the lucky or lucky CLI commit so we remember to note this in the upgrade guides

PR #466, opened 2020-03-14
Native Task params

### LINKED ISSUE #378: Native support for task args
Right now you can use `ARGV`, but that feels like a "loop hole" more than implementation. It'd be nice if tasks had a common interface for when you want pass args in to your task like `lucky gen.model User`.

Maybe we could have a wrapper around OptionParser? 

```crystal
class Gen::Model < LuckyCli::Task
  summary "generate a model" 
  # lucky gen.model -m User
  # lucky gen.model --model User
  # lucky gen.model User
  # lucky gen.model user  (error arg user doesn't match format...)
  param model_name, as: ["-m", "--model", ""], format: /^[A-Z].+/

  def call
    @model_name
  end
end
```

With something like this, we could add some options to it for error handling, validations, and allowing for passing in multiple args easily. 

ref: https://gitter.im/luckyframework/Lobby?at=5d4e0ca49507182477af244b

jwoertink:
Fixes #378

This PR adds built-in task args which would allow you to do things like these

```
lucky search.reindex User
lucky users.expired -f
lucky generate_sitemaps --environment=DEVELOPMENT
```

```crystal
class Search::Reindex < LuckyCli::Task
  summary "Reindex records for a model"

  positional_arg :model, "Specify which model to reindex", required: true

  def call
     # made up code to reindex User
     Elasticsearch.reindex(model: model)
  end
end

class Users::Expired < LuckyCli::Task
  summary "Expire users"
  switch :force, "force expire users", shortcut: "-f"

  def call
    users = UserQuery.new.expired

     users.each do |user|
        user.expire! if force?
     end
  end
end

class GenerateSitemaps < LuckyCli::Task
  summary "Generate some sitemaps"
  arg :environment, "Specify the ENV for the sitemaps", format: /(dev|prod)/

  def call
    # raises an error because it expects `dev` but `DEVELOPMENT` was passed in
     Sitemapper.generate(env: environment)
   end
end
```


src/lucky_cli/task.cr
```diff
@@ -4,6 +4,11 @@ abstract class LuckyCli::Task
       LuckyCli::Runner.tasks << self.new
     {% end %}
 
+    private getter params : Array(String) = [] of String
+
+    annotation ParamFormat
```
jwoertink: I'm not sold on this name, but it's a start

src/lucky_cli/task.cr
```diff
@@ -20,12 +25,17 @@ abstract class LuckyCli::Task
...
         io.puts help_message
       else
+        parse_and_store_args(args)
         call
       end
     end
 
     private def wants_help_message?(args)
-      args.any? { |arg| ["--help", "-h", "help"].includes?(arg) }
+      args.any? { |arg| {"--help", "-h", "help"}.includes?(arg) }
```
jwoertink: This is just a small optimization win.

src/lucky_cli/runner.cr
```diff
@@ -14,7 +14,7 @@ class LuckyCli::Runner
   def self.run(args = ARGV, io : IO = STDERR)
     task_name = args.shift?
 
-    if !task_name.nil? && ["--help", "-h"].includes?(task_name)
+    if !task_name.nil? && {"--help", "-h"}.includes?(task_name)
```
jwoertink: Small optimization win

spec/support/tasks.cr
```diff
@@ -24,3 +24,35 @@ class AnotherTask < LuckyCli::Task
   def call
   end
 end
+
+class TaskWithParam < LuckyCli::Task
+  summary "This task has a param"
+  param model_name
```
jwoertink: No type added because a param will always be a string. If the user needs a number then they can parse and typecast as needed.

paulcsmith:
I'm way excited about this. Having a built in way to handle args will be sooooo nice. I've been thinking about this a bit and I think there are a couple things we should also cover, and which may mean the design of this would have to change as well

* **Positional args**. Stuff like the lucky generators don't have an arg name, just position. Like the model/page/component generators that look for the first passed in arg
* **Switches**. Simple true/false flags like `--no-api` or `--full`
* **Format like other CLIs**. `--model-name=User` or with a shortcut `-m User`

I think to get the parsing for switches and CLI formatting we could use the built-in [OptionParser](https://crystal-lang.org/api/0.33.0/OptionParser.html)

I'm imagining that the macros could generate methods that use the option parser under the hood...but I dunno exactly how it'd work. In terms of what the user sees maybe it'd be something like this

```class
class MyTask < LuckyCli::Task
  summary "Do some awesome stuff"

  positional_arg :user_email, format: /some-regex/, required: true
  positional_arg :user_type, in: ["foo", "bar"], required: true
  switch :admin, shortcut: "-a" # Looks for a switch with --admin or -a
  arg :nickname, shortcut: "-n" # Looks for an arg '--nickname=NICKNAME' or '-n NICKNAME'
  
  def call
  end
end
```

A total pseudo code implementation of how the option parser might work. This may be a dumb idea, but thought I'd put it out there.

```
class LuckyCli::Task
  ARG_INFO = {positional_args: 0}

  macro arg(arg_name, shortcut = nil, required = false, format = nil, in = nil)
    @{{ arg_name.id }} : String?

    def {{ arg_name.id }} : String
      # Format and in can be handled before this somehow
      OptionParser.parse do |parser|
        parser.on("-#{shortcut} {{ arg_name.stringify.upcase }}", "--{{ arg_name.stringify.under_score.gsub("_", "-")={{ arg_name.stringify.upcase }}") do |value|
          @{{ arg_name.id }} = value
        end
      end 
      # If requires then make sure it isn't nil
      @{{ arg_name.id }}.not_nil!
    end
  end

  macro switch(arg_name, shortcut = nil)
    @{{ arg_name.id }} : Bool = false

    def {{ arg_name.id }}? : Bool
      OptionParser.parse do |parser|
        parser.on("-#{shortcut}", "--{{ arg_name.stringify.underscore.gsub("_", "-") }}") do
          @{{ arg_name.id }} = true
        end
      end
        @{{ arg_name.id }}
    end
  end

  macro positional_arg(arg_name) # Just doing arg_name for conciseness
    @{{ arg_name.id }} : String

    def {{ arg_name.id }} : String
      ARGV[{{  ARG_INFO[:positional_args] }}]
      {% ARG_INFO[:positional_args] = ARG_INFO[:positional_args] + 1 %}
    end
  end
end
```

What do you think of this?

jwoertink:
That seems like a lot of different options to do the same thing a bunch of different ways. What if we just changed the API of how our generators worked? Like always require a flag instead of positional arguments. That makes `arg` and `switch` the same, then we just have 1 method for defining that. I like the name `arg` for that.

```
lucky gen.model -m Project -c title:String -c completed:Bool
lucky gen.page --name Users::IndexPage
lucky db.schema.dump --file structure.sql
lucky custom.init --no-api
```

I know this isn't pretty, but it does create consistency across all of the tasks. 


OR ... just had a thought...


```crystal
class Gen::Model < LuckyCli::Task
  arg :model, shortcut: 0
  # a little funky, but something representing everything after the first one
  arg :columns, shortcut: 1..-1
end
```

The idea here being that we can use flags, but as a shortcut, we just put the args in the order specified... 🤷‍♂

paulcsmith:
I feel pretty strongly that we keep positional args in our generators just because it's a little simpler and I think more expected. Also a lot of CLI tools have positional args like git, Rails, etc.

But I agree that is a lot of methods. 

Could you explain the how `switch` wouldn't be necessary though? Seems like it still would since it is a boolean so requires no `=` or value.

Maybe a hybrid of our approaches would work?

```crystal
arg :model_name, position: 0
arg :type, shortcut: "-t"
switch :force, shortcut: "-f"
```

That way there is just arg and switch. Where switch is always a bool that defaults to off. And arg can be made positional with the `position` arg. Thoughts?

jwoertink:
Yeah, that's not a bad solution. A few things though. 

* How do we handle `--no-api`? `switch :"no-api"` or `switch :no_api`?
* How do we tell positional that there's more than one in the case of model columns? `arg :columns, positional: 1..`?
* Not sure, but does OptionParser let you choose between `--name=Jeremy` and `--name Jeremy`?
* Would it be cleaner to just add more options like `arg :no_api, switch: true`? or does that just make it more confusing vs having multiple methods to do it?

paulcsmith:
> - How do we handle --no-api? switch :"no-api" or switch :no_api?

I think you'd always use a symbol and we would change it to dashed

> - How do we tell positional that there's more than one in the case of model columns? arg :columns, positional: 1..?

Hmm good point! Maybe we change it to `array_from: 1` or something like that?

> - Not sure, but does OptionParser let you choose between --name=Jeremy and --name Jeremy?

I'm not sure, but I think we could choose one and make that the default. My personal preference is using `=` since that's what I've seen and I think it makes more sense but as long as we have one and its documented I think that's ok

> - Would it be cleaner to just add more options like arg :no_api, switch: true? or does that just make it more confusing vs having multiple methods to do it?

I think switch seems like a special case that is different enough in both return type (and options) that it makes sense to be its own thing. For example a switch doesn't need `in`, `format` or really any kind of validations, so it seems like its own thing to me

jwoertink:
Ok. I'll give that a shot and push up some changes. I can try and get to that today, but if you wanted to give it a shot, I would not be offended if you closed this 😅

paulcsmith:
@jwoertink If you want to do it that'd be great! I'm working on some other stuff on Avram today! I'm happy we talked this through and arrived at a good hybrid approach. I think this will make building tasks soooo much easier, and will remove a lot of custom code from our generators 💚

jwoertink:
@paulcsmith so it turns out that `OptionParser` requires a help description (which makes sense)... so what do you think that looks like?

```crystal
arg :model_name, description: "Defines which model to index", required: true, format: //, positional: 0
switch :force, shortcut: "-f", description: "Force drop and overwrite..."
```

or do we attempt to just put something generic?

paulcsmith:
Hmm I think requiring a description would be a good idea. Then we can add that to the default help text in a later PR :D

wontruefree:
Are all params strings? and if you want to enable other types would this change limit this in the future?

jwoertink:
@wontruefree yeah, all params would be String. It would be up to you to handle that parsing if you wanted something different. 

I had thought about adding in types originally, but it becomes a lot harder to support more. Like supporting Hash, or UInt... We could probably do String, Int32, Float64, and maybe Array of those things, but beyond that I think it becomes way too complicated.

paulcsmith:
Yeah I agree. Can do something like this in the meantime

```
def number_of_pages
  previous_def.to_i
end
```

In the future maybe we could limit casting to other types to simple types like Int16/32/64 and Time. But this work shouldn't make that any harder so can be done later 👍

jwoertink:
Ok, I've updated the original post. There's lots going on here, but the main thing is I had to roll with 3 different methods to achieve this. The positional args aren't supported natively in the crystal OptionParser from what I could tell, so they have to be handled completely different. Also since positional args don't need a `shortcut`. 

One thing that is sort of wonky at the moment is how the help menu is handled. The positional_arg takes a description, but it's not passed to the `OptionParser`, so you don't get to see what that description is. Maybe it doesn't need it? Or maybe we can add it later when we clean up how the help menu works? 

Overall this is at least a start, but will need some cleanup love for sure.

spec/lucky_cli/runner_spec.cr
```diff
@@ -4,11 +4,7 @@ include HaveDefaultHelperMessageMatcher
 
 describe LuckyCli::Runner do
   it "adds tasks to the runner when task classes are created" do
-    LuckyCli::Runner.tasks.map(&.name).reject(&.==("lucky_cli.dev")).should eq [
-      "another_task",
-      "my.cool_task",
-      "my.custom_name",
-    ]
+    LuckyCli::Runner.tasks.map(&.name).reject(&.==("lucky_cli.dev")).should contain("another_task")
```
jwoertink: This was just because this list kept getting longer, and failing.

src/lucky_cli/task.cr
```diff
@@ -60,6 +68,131 @@ abstract class LuckyCli::Task
     end
   end
 
+  macro positional_arg(arg_name, description, required = nil, format = nil, to = nil)
+    {% PARSER_OPTS << arg_name %}
+    @{{ arg_name.id }} : {% if to %}Array(String){% else %}String{% end %} | Nil
```
jwoertink: Positional args could be a `String` like `User`, but also `Array(String)` like `["name:String", "email:String"]`.

src/lucky_cli/task.cr
```diff
@@ -60,6 +68,131 @@ abstract class LuckyCli::Task
...
+        @{{ arg_name.id }}
+      {% end %}
+    end
+  end
+
+  macro arg(arg_name, description, shortcut = nil, required = nil, format = nil)
+    {% PARSER_OPTS << arg_name %}
+    @{{ arg_name.id }} : String?
+
+    def set_opt_for_{{ arg_name.id }}(unused_args : Array(String))
```
jwoertink: I added this because positional_arg requires passing in the args, and I wanted a consistent interface so I didn't have to do some context switching to figure out if I pass an arg or not.

src/lucky_cli/task.cr
```diff
@@ -60,6 +68,131 @@ abstract class LuckyCli::Task
...
+        @{{ arg_name.id }}
+      {% end %}
+    end
+  end
+
+  macro switch(arg_name, description, shortcut = nil)
+    {% PARSER_OPTS << arg_name %}
+    @{{ arg_name.id }} : Bool = false
+
+    def set_opt_for_{{ arg_name.id }}(unused_args : Array(String))
```
jwoertink: same as `arg`.

paulcsmith:
There is a lot here and I'm feeling spread thin so will take a look later! Sorry it'll take a bit longer

jwoertink:
@paulcsmith hang in there dude ❤️  No need to rush on any of this.

src/lucky_cli/task.cr
```diff
@@ -60,6 +68,131 @@ abstract class LuckyCli::Task
...
+        if value =~ {{ format }}
+          @{{ arg_name.id }} = value
+        else
+          raise "Invalid format for {{ arg_name.id }}. It should match {{ format }}"
+        end
+        {% else %}
+          @{{ arg_name.id }} = value
+        {% end %}
+      end
+      {% end %}
```
bdtomlin: I noticed both option parsers had the same body, so maybe they could be dried up a little.
```suggestion
      option_parser.on(
        {% if shortcut %}"{{ shortcut.id }} {{ arg_name.stringify.upcase.id }}",{% end %}
        "--{{ arg_name.id.stringify.underscore.gsub(/_/, "-").id }}={{ arg_name.id.stringify.upcase.id }}",
        {{ description }}
      ) do |value|
        value = value.strip
        {% if format %}
        if value =~ {{ format }}
          @{{ arg_name.id }} = value
        else
          raise "Invalid format for {{ arg_name.id }}. It should match {{ format }}"
        end
        {% else %}
          @{{ arg_name.id }} = value
        {% end %}
      end
```

src/lucky_cli/task.cr
```diff
@@ -60,6 +68,131 @@ abstract class LuckyCli::Task
...
+    end
+
+    def {{ arg_name.id }} : String{% if !required %}?{% end %}
+      {% if required %}
```
bdtomlin: Same comment as above... wondering if the required validation and exception should run when setting the value instead of reading it. I didn't do a code example because it would muddy the previous comment about drying up `set_opt_for_...`.

src/lucky_cli/task.cr
```diff
@@ -60,6 +68,131 @@ abstract class LuckyCli::Task
...
+        if @{{ arg_name.id }}.nil?
+          raise <<-ERROR
+          {{ arg_name.id }} is required, but no value was passed.
+          ERROR
+        end
+        @{{ arg_name.id }}.not_nil!
+      {% else %}
+        @{{ arg_name.id }}
+      {% end %}
+    end
```
bdtomlin: Do you think `required` should be checked in `set_opt_for_...`? That makes a little more sense to me than waiting until you try to access the method. This way you would not be allowing bad data in, and it would consolidate the validations (format and required).
```suggestion
    def set_opt_for_{{ arg_name.id }}(args : Array(String))
      {% if to_end %}
        value = args[@positional_arg_count..-1]
      {% else %}
        value = args[@positional_arg_count]?
      {% end %}
      {% if required %}
        if value.nil?
          raise <<-ERROR
          {{ arg_name.id }} is required, but no value was passed.
          ERROR
        end
      {% end %}
      {% if format %}
      matches = value.is_a?(Array) ? value.all?(&.=~({{ format }})) : value =~ {{ format }}
      if !matches
        raise "Invalid format for {{ arg_name.id }}. It should match {{ format }}"
      end
      {% end %}
      @{{ arg_name.id }} = value
      @positional_arg_count += 1
    end
    def {{ arg_name.id }}
      @{{ arg_name.id }}{% if required %}.not_nil!{% end %}
    end
```

spec/lucky_cli/runner_spec.cr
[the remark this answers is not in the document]
jwoertink: ah, good idea. This looks like a nice middle ground 👍

src/lucky_cli/task.cr
```diff
@@ -60,6 +68,131 @@ abstract class LuckyCli::Task
     end
   end
 
+  macro positional_arg(arg_name, description, required = nil, format = nil, to = nil)
```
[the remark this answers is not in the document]
jwoertink: My original thought was actually to just use a range object so you could grab any group like `[1..]` or `[1..4]`, but I figured that might be a bit difficult, and confusing. 

I like your ideas here, I'll give that a shot.

src/lucky_cli/task.cr
jwoertink: Good call 👍

src/lucky_cli/task.cr
jwoertink: I wanted to do this, but ran in to an issue. Maybe you can help me through it though? For `arg`, you don't know if the value is set until after the options have been parsed. The parsed method isn't called until after they've already been set. For example:

```crystal
@val = nil
parser.on("--x=X", "") do |val|
  @val = val
  #if you get here, `val` has a value
end
@val == nil
parser.parse(ARGV)
# now you know if @val is nil or not, but the validation check would have to be here
```

This would mean I'd have to add a new method to run post parsing validations (which I'm doing now, but just in the form of you calling those methods). Maybe that's ok? 🤷‍♂

src/lucky_cli/task.cr
jwoertink: I agree, it would be better to check sooner. On positional args this is possible, but on flagged args, it's a little more difficult. I wanted it to be consistent so you weren't guessing when it would raise an error and when it wouldn't. I left a comment on the other one about the issue I ran in to, so if you have an idea for how to get around that, let me know.

jwoertink:
@bdtomlin Thanks for reviewing! I made most of those changes. I think this is a lot easier to consume when reading over now. I also made a few other small adjustments like comments and such.

src/lucky_cli/task.cr
```diff
@@ -60,6 +68,129 @@ abstract class LuckyCli::Task
...
+      else
+        raise "Invalid format for {{ arg_name.id }}. It should match {{ format }}"
+      end
+      {% else %}
+        @{{ arg_name.id }} = value
+      {% end %}
+      @positional_arg_count += 1
+    end
+
+    def {{ arg_name.id }}
```
jwoertink: I wanted to have the return type set, but there's 4 possible options here `Array(String)`, `Array(String)?` `String` and `String?` which makes the macro code really funky and difficult to read. I figure we can let the compiler handle it on this case.

src/lucky_cli/task.cr
bdtomlin: Yeah, I think it's ok. I had looked at the positional_arg macro and it works their, but I hadn't considered in the arg macro it wasn't possible. I jumped through a few mental contortions to try to figure it out, but it seems to work well as it is. 😄

src/lucky_cli/task.cr
```diff
@@ -60,6 +68,129 @@ abstract class LuckyCli::Task
...
+        value = value.strip
+        {% if format %}
+        if value =~ {{ format }}
+          @{{ arg_name.id }} = value
+        else
+          raise "Invalid format for {{ arg_name.id }}. It should match {{ format }}"
+        end
+        {% else %}
+          @{{ arg_name.id }} = value
+        {% end %}
```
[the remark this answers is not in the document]
jwoertink: I didn't know `!~` was a thing 😅

bdtomlin:
@jwoertink thanks for working with me on this. I saw this PR was a little complicated and daunting to get started with, so I wanted to give another set of eyes for everyone to maybe help reduce the stress.

It's hard for me to start doing code reviews on a new code base because I don't like being critical, especially when the code is already better than I would have written. 😅

Thanks for making it easy to get involved.

spec/lucky_cli/runner_spec.cr
```diff
@@ -4,11 +4,13 @@ include HaveDefaultHelperMessageMatcher
 
 describe LuckyCli::Runner do
   it "adds tasks to the runner when task classes are created" do
-    LuckyCli::Runner.tasks.map(&.name).reject(&.==("lucky_cli.dev")).should eq [
-      "another_task",
-      "my.cool_task",
-      "my.custom_name",
-    ]
+    expected_task_names = ["another_task", "my.cool.task", "my.custom_name"]
```
bdtomlin: This is why the tests are failing
```suggestion
    expected_task_names = ["another_task", "my.cool_task", "my.custom_name"]
```

jwoertink: ha! Good catch 😅  For some reason all of the specs fail locally on my machine, so not sure which ones are "real" errors and which are just my machine being weird 😕  I should take some time to figure that out.

bdtomlin:
I have to run specific specs by themselves on my machine. The init_web_spec is the issue. I believe it causes my system to run out of memory and that causes the failures. I have 16GB of member :[

jwoertink:
Interesting. I wonder if mine is the same deal. I'd have to look, but maybe we're booting flow in there somewhere, and not closing the chromedriver? I hear memory being eaten, and I think "CHROOOMMMEEE!!!!" ✊  😂

src/lucky_cli/task.cr
```diff
@@ -60,6 +68,123 @@ abstract class LuckyCli::Task
...
+        raise "Invalid format for {{ arg_name.id }}. It should match {{ format }}"
+      end
+      {% end %}
+      @{{ arg_name.id }} = value
+      @positional_arg_count += 1
+    end
+
+    def {{ arg_name.id }}
+      {% if required %}
+        if @{{ arg_name.id }}.nil?
```
paulcsmith, line 105: I'm kinda thinking that `positional_arg` that aren't `to_end` should always be required. Otherwise if you have 3 args and the middle one is optional should the second arg be mapped to the third argument? Or did they just forget the second arg?

So I'm thinking we require positional_args

For positonal_args with `to_end` I think it can be optional by checking `empty?` So by default it requires *something* but if set to optional it is ok with an empty array. This would also need a matching spec of course

Thoughts?

src/lucky_cli/task.cr
paulcsmith: I agree it'd be nice and I'm also ok shipping as -is. Maybe we come up with something later!

src/lucky_cli/task.cr
```diff
@@ -60,6 +68,123 @@ abstract class LuckyCli::Task
...
+      end
+    end
+
+    def {{ arg_name.id }} : String{% if !required %}?{% end %}
+      {% if required %}
+        if @{{ arg_name.id }}.nil?
+          raise <<-ERROR
+          {{ arg_name.id }} is required, but no value was passed.
+
+          Try this...
```
paulcsmith, line 149: Gotta love a good error message! :D

src/lucky_cli/task.cr
```diff
@@ -60,6 +68,123 @@ abstract class LuckyCli::Task
...
+  #
+  # If your arg takes more than one value, you can set `to_end` to true to capture all
+  # args from this position to the end. This will make your `arg_name` method return `Array(String)`.
+  #
+  # * `arg_name` : String - The name of the argument
+  # * `description` : String - The help text description for this option
+  # * `required` : Bool - raise exception if this arg is not passed
+  # * `to_end` : Bool - Capture all args from this position to the end.
+  # * `format` : Regex - The format you expect the args to match
+  macro positional_arg(arg_name, description, required = false, to_end = false, format = nil)
```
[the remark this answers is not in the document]
jwoertink: I'm cool with that.

spec/support/tasks.cr
```diff
@@ -24,3 +24,44 @@ class AnotherTask < LuckyCli::Task
...
+  def call
+    self
+  end
+end
+
+class TaskWithPositionalArgs < LuckyCli::Task
+  summary "This is a task with positional args"
+
+  positional_arg :model, "Define the model", required: true, format: /^[A-Z]/
+  positional_arg :columns, "Define the columns like name:String", to_end: true, format: /\w+:[A-Z]\w+(::\w+)?/
```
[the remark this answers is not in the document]
jwoertink: `positional_arg` and the others don't play well. I would need to know the position of the args that aren't positional to know which index to start at. Or the positional arg would have to know the existence of the others in order to gsub out those..

```crystal
  arg :template, "Add which template you want to use", shortcut: "-t"
  switch :add_auth, "Add in auth", shortcut: "-a"
  positional_arg :options, "Add which options you need", to_end: true, format: /\w+/
```

Right now the `positional_arg` index is 0. That means in this:

`-t BLAH -a one two three`

`template` would be "BLAH", `add_auth` would be true, and `options` would be `["-t BLAH", "-a", "one two three"]`. This would be the case regardless of the order. It's mainly because `arg` and `switch` use the built-in option parser, but `positional_arg` has to hand roll it.

I could do some dirty hacks like tracking if arg or switch are used. In this example it could work, but if you put positonal_arg second, and switch after, but positional_arg said "to_end", it would bork again. Maybe that's just like "hey, you're the one that put grape jelly on your roast beef, you deal with it"..... on a side note, I bet strawberry jam on a turkey sandwich would be bomb... 🤔

spec/lucky_cli/task_spec.cr
```diff
@@ -39,4 +39,49 @@ describe LuckyCli::Task do
       io.to_s.chomp.should eq "Custom help message"
     end
   end
```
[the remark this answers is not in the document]
jwoertink: This will be tricky with the positional_arg. The others pass description to the built-in option parser, but since the positional has to be hand rolled, there's no place the description is actually used. 

My thought here is make a constant and push all of the data in to that, then iterate over this to display our own custom arg menu deal. This means that I need to know on all of them if there's a shortcut or not, which type it is, maybe the format? So this ends up looking like:

```crystal
{long: "--model_name=NAME", short: "-m NAME", format: "/\w+/", desc: "The name of the model"}
{long: "--force", short: "", format: "", desc: "Force this to work"}
{long: "Position 0", short: "", format: "/\w+/", desc: "First positional arg thing"}
{long: "Position 1..-1", short: "", format: "/\w+/", desc: "First positional arg thing"}
```

I don't really know how to best show the format for the positional, but they should all be the same so then I can do something like

```
CLI_ARG_HELP_OPTS.each do |data|
  str << data[:long]
  str << data[:short] if data[:short]
  str << "some sort of padding...." data[:desc]
  str << "(Format: #{data[:format]})"
end
```

spec/support/tasks.cr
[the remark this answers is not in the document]
paulcsmith: First off, stop making me hungry 😂 

Second yeah this is complex. Maybe we  just forget it for now and maybe we get ideas come later

This is way better than what we have today so I’m cool with shipping as is!

spec/lucky_cli/task_spec.cr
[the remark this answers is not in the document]
paulcsmith: Hmm seems pretty complex. If you want to do that then I’m down but I’m also ok shipping as is. This is so awesome that I wouldn’t want to hold it up even longer haha. 

So up totally up to you!

spec/lucky_cli/task_spec.cr
```diff
@@ -39,4 +39,49 @@ describe LuckyCli::Task do
...
+      task = TaskWithSwitchFlags.new.print_help_or_call(args: ["-a"]).not_nil!
+      task.force?.should eq false
+      task.admin?.should eq true
+    end
+
+    it "allows the args to be positional with no flags" do
+      task = TaskWithPositionalArgs.new.print_help_or_call(args: ["User", "name:String", "email:String"]).not_nil!
+      task.model.should eq "User"
+      task.columns.should eq ["name:String", "email:String"]
+    end
```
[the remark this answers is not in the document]
paulcsmith: Or I guess this won’t work so maybe just switch and args?

jwoertink:
Ok, there's more I'd like to do with this, but it's a bit complicated. I think we just get this out there in a v1 type status, play with it, and see how we feel about how it works. I'm sure we can hammer out some refactors to make them play nicely together, and better documentation and such in a future rollout. 

If this is all green, I'll merge it in.

paulcsmith:
Sounds good and agree on shipping early and seeing how it goes. I think this is a big stpe forward and good to keep momentum 👍

PR #469, opened 2020-03-17
adding in a generated default DB port.

jwoertink:
In the PR https://github.com/luckyframework/avram/pull/333 we are removing the default port value to fix some other issues with connecting to sockets. This PR adds in the port default to your new generated lucky app.


jwoertink:
We were force specifying the default port before, but I'm pretty sure postgres will just fallback to it's default without specifying anyway, so this change isn't totally required. This just moves that setting down to the user level so people have the option to not use it if they need. Otherwise this should all work as it does now.

paulcsmith:
Ah ok so people can leave it off and Postgres will just the default port if not given?

If so that sounds good to me!

jwoertink:
Yeah, passing in the port isn't necessary if postgres is running on the default port. And really we're just talking about this update for development, so production isn't affected at all. The main benefit here is just that if someone needs to change the port locally, it's a lot more apparent how and where to do that. (or if using docker for development)

PR #472, opened 2020-03-18
Add permissive robots.txt

### LINKED ISSUE #470: Add a default robots.txt to /public
Just an idea. I could PR a permissive one really fast.

watzon:
Closes #470

### DIFF
--- a/src/browser_app_skeleton/public/robots.txt
+++ b/src/browser_app_skeleton/public/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Allow: *
--- a/src/browser_app_skeleton/public/robots.txt
+++ b/src/browser_app_skeleton/public/robots.txt
@@ -1,2 +1,3 @@
+# https://www.robotstxt.org/robotstxt.html
 User-agent: *
 Disallow:
--- a/src/browser_app_skeleton/public/robots.txt
+++ b/src/browser_app_skeleton/public/robots.txt
@@ -1,2 +1,2 @@
 User-agent: *
-Allow: *
+Disallow:


src/browser_app_skeleton/public/robots.txt
```diff
@@ -0,0 +1,2 @@
+User-agent: *
+Allow: *
```
paulcsmith: I've done a bit of research and it looks like `Disallow` is better supported

```suggestion
Disallow:
```

Thoughts on changing to that?

src/browser_app_skeleton/public/robots.txt
```diff
@@ -0,0 +1,2 @@
+User-agent: *
```
jwoertink: Can you add a comment just above this pointing to https://www.robotstxt.org/robotstxt.html ? Just a quick little reference saying "For info on updating this" or whatever.

src/browser_app_skeleton/public/robots.txt
wontruefree: That seems confusing.  Maybe we can add a comment.

paulcsmith: Yeah it is super weird. Comment sounds good 👍

watzon: Yeah really it shouldn't need either afaik, but changing to disallow could work fine

src/browser_app_skeleton/public/robots.txt
watzon: Done :+1:

paulcsmith:
Thank you!

PR #481, opened 2020-03-20
Add compiling message

### LINKED ISSUE #480: Add a "Compiling xxx.task" if there is no precompiled version
Sometimes it looks frozen so showing a message would be great. No message needs to be shown if precompiled since they are instant.

Would also LOVE to figure out a way to get some kind of loading thing like oclif, but that can be added later

paulcsmith:
Closes #480

compile-spinner

### DIFF
--- a/src/build_and_run_task.cr
+++ b/src/build_and_run_task.cr
@@ -1,0 +1,60 @@
+# Builds and runs a file (typically tasks.cr) with the passed in args
+class LuckyCli::BuildAndRunTask
+  private getter tasks_file : String
+  private getter args : String
+  private getter temp_stdout = IO::Memory.new
+  private getter temp_stderr = IO::Memory.new
+  private getter tasks_binary_path = "tmp/tasks_binary"
+
+  def initialize(@tasks_file, @args)
+  end
+
+  def self.call(*args, **named_args)
+    new(*args, **named_args).call
+  end
+
+  def call
+    FileUtils.mkdir_p("tmp")
+    build_status = build_tasks_binary
+    copy_temp_io_to_real_io
+
+    if build_status.success?
+      run_tasks_file
+    else
+      exit build_status.exit_code
+    end
+  end
+
+  private def build_tasks_binary
+    LuckyCli::Spinner.start("Compiling...") do
+      Process.run(
+        "crystal build #{tasks_file} -o #{tasks_binary_path} ",
+        shell: true,
+        input: STDIN,
+        output: temp_stdout,
+        error: temp_stderr
+      )
+    end
+  end
+
+  # STDOUT/STDERR needs to be printed after the tasks binary is built,
+  # otherwise the spinner will overwrite the output from building the binary
+  # (like if there is a compilation error when building).
+  #
+  # By creating "fake" io with IO::Memory we can collect the output and then
+  # print it after the build is finished.
+  private def copy_temp_io_to_real_io
+    STDOUT.print(temp_stdout.to_s)
+    STDERR.print(temp_stderr.to_s)
+  end
+
+  private def run_tasks_file
+    exit Process.run(
+      "#{tasks_binary_path} #{args}",
+      shell: true,
+      input: STDIN,
+      output: STDOUT,
+      error: STDERR
+    ).exit_code
+  end
+end
--- a/src/lucky_cli/spinner.cr
+++ b/src/lucky_cli/spinner.cr
@@ -1,0 +1,130 @@
+require "colorize"
+
+class LuckyCli::Spinner
+  DELETE_LINE = "\33[2K\r"
+
+  private property frame_index : Int32 = 0
+  private property? started : Bool = false
+  private property last_frame_rendered_at : Time? = nil
+
+  @done_text : String | ColoredString | Nil
+  @start_text : String | ColoredString
+  getter colorize_frame : Proc(String, ColoredString)
+  getter io : IO
+
+  alias ColoredString = Colorize::Object(String)
+
+  def initialize(
+    @start_text,
+    @io = STDERR,
+    @done_text : String | ColoredString | Nil = nil,
+    @colorize_frame = ->(frame : String) { frame.colorize.bold.green }
+  )
+  end
+
+  def self.start(*args, **named_args)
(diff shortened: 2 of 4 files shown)


src/build_and_run_task.cr
```diff
@@ -0,0 +1,60 @@
...
+  # STDOUT/STDERR needs to be printed after the tasks binary is built,
+  # otherwise the spinner will overwrite the output from building the binary
+  # (like if there is a compilation error when building).
+  #
+  # By creating "fake" io with IO::Memory we can collect the output and then
+  # print it after the build is finished.
+  private def copy_fake_io_to_real_io
+    STDOUT.print(fake_stdout.to_s)
+    STDERR.print(fake_stderr.to_s)
+  end
```
paulcsmith: This isn't ideal but works fine for our use case. Ideally the spinner would have some logic to print stdout/err above the spinner line, but that can be added later

src/lucky_cli/spinner.cr
```diff
@@ -0,0 +1,134 @@
...
+class LuckyCli::Spinner
+  DELETE_LINE = "\33[2K\r"
+
+  private property frame_index : Int32 = 0
+  private property? started : Bool = false
+  private property last_frame_rendered_at : Time? = nil
+
+  @done_text : String | ColoredString | Nil
+  @start_text : String | ColoredString
+  getter colorize_frame : Proc(String, ColoredString)
```
paulcsmith: Allow customizing the color of the frame

src/lucky_cli/spinner.cr
```diff
@@ -0,0 +1,134 @@
+require "colorize"
+
+class LuckyCli::Spinner
+  DELETE_LINE = "\33[2K\r"
+
+  private property frame_index : Int32 = 0
+  private property? started : Bool = false
+  private property last_frame_rendered_at : Time? = nil
+
+  @done_text : String | ColoredString | Nil
```
paulcsmith: By default this deletes the line, but you can provide "done_text" to replace the spinner part with text: `Compiling... done`

src/lucky_cli/spinner.cr
```diff
@@ -0,0 +1,134 @@
...
+    self.started = true
+    print_spinner
+
+    spawn do
+      while started?
+        if ready_for_next_frame?
+          increment_frame
+          print_spinner
+        end
+        Fiber.yield
```
paulcsmith: Need this so the current fiber has a chance to run its code. Otherwise this while just blocks the scheduler forever. This took me awhile to figure out :P

src/build_and_run_task.cr
jwoertink: I think the process makes sense. Maybe we can call it `temp_stdout` or something? Using `fake` seems weird... but I guess we can just roll with it for now. We've been known to change up names of stuff later 😅

src/lucky_cli/spinner.cr
```diff
@@ -0,0 +1,130 @@
...
+    [
+      "⣾",
+      "⣽",
+      "⣻",
+      "⢿",
+      "⡿",
+      "⣟",
+      "⣯",
+      "⣷",
+    ]
```
jwoertink: Maybe an irrelevant optimization, but this could be a Tuple with Chars.

src/build_and_run_task.cr
```diff
@@ -0,0 +1,60 @@
...
+  def initialize(@tasks_file, @args)
+  end
+
+  def self.call(*args, **named_args)
+    new(*args, **named_args).call
+  end
+
+  def call
+    FileUtils.mkdir_p("tmp")
+    build_status = build_tasks_binary
```
jwoertink: I wonder if we could do some sort of cache here. Maybe compare md5 or something so we don't have to build every time. 🤔  That could be a future thing, but that would be awesome if the second time you ran a task it was a lot faster.

src/build_and_run_task.cr
paulcsmith: Good idea! I like `temp_stdout/err`

src/lucky_cli/spinner.cr
paulcsmith: Might as well!

src/build_and_run_task.cr
paulcsmith: Yes I think so. I was thinking about how to do it and I think we'd need to extract/copy some of the code in lucky watch that checks the timestamps. Then we'd compare the timestamps of the tasks binary and the last modifification date of the built files, but I'm not sure. Would definitely be cool so if you run a few tasks in a row you would only compile once!

I'll do it in a separate change at some point though since caching stuff is known to be quite tricky :P

src/lucky_cli/spinner.cr
paulcsmith: Actually I'm going to keep it as an array, but make it a class getter so it is intiialized once. Doing it with a Tuple was a pain because I need to do `Tuple(String, String, String, String, :P)`

src/build_and_run_task.cr
jwoertink: Yeah, a separate PR for sure. This is already quite a bit. But I just had an idea. Before compiling, we just take the file before we build it, and get the sha and/or timestamp of it, and then create a lucky cache directory. Then we just check the cache and skip building. There's already a `~/.cache/crystal/` and `~/.cache/shards/`. We might as well just hook in to that.

paulcsmith: @jwoertink that is a cool idea. I was thinking of how it’d work but I’m not sure it will. We could cache the sha of the tasks file, but the tasks file depends on a bunch of other files and folders, like models, tasks from shards, etc so it’d have to track those too. So we’d have to sha everything which I don’t think would work. 

Even if we could sha everything  I think it may be simpler to do a last modified traversal where we give it a list of files and folders and it gets the last modified file or folder and compares with the last modified timestamp on the binary. I think that’d work since if any files change, like a model, it’ll recompile the tasks binary. Does that make sense or is there another way to approach this?

paulcsmith:
Or possible the JsonBodyParser https://github.com/luckyframework/lucky/pull/1053/files#diff-d633a5eeaf04a8d42861bcc3614b84bfR477

PR #482, opened 2020-03-21
Refactor bash scripts with some new updates

jwoertink:
Fixes https://github.com/luckyframework/lucky/issues/953

This has evolved a bit from it's first iteration. This PR refactors the setup script a bit and adds in some new stuff.

* Adds new `system_check` script which is added to your `Procfile.dev`. This can check for things that need to run before your app is booted each time.
* Moves the setup checks from `setup` to `system_check`. Now `setup` just installs what it needs.
* Added some abstractions to clean up the `setup` script which moves those abstractions to a new helper directory
* Added 2 new `text_helpers` and `function_helpers` to make writing bash easier because let's face it.. it's hard


src/web_app_skeleton/Procfile.dev.ecr
```diff
@@ -1,3 +1,4 @@
+system_check: ./script/system_check
```
jwoertink: Run it for both API and web because you may have external deps in both cases.

src/web_app_skeleton/script/system_check
```diff
@@ -0,0 +1,31 @@
...
+# A `print_error` function is provided as a helpful way to display a custom
+# error message, and exit the script when a process fails, or is missing.
+#
+# A few examples might be:
+#   * `lucky db.verify_connection` to test postgres can be connected
+#   * Checking for elasticsearch, redis, or any other process is booted first
+#   * Booting additional processes for things like mail, background jobs, etc...
+
+# run command here
+# if [ $? -ne 0 ]
```
jwoertink: If the exit code is not equal to 0, then it's an error. https://shapeshed.com/unix-exit-codes/

This seems to be fine for zsh and bash.

src/web_app_skeleton/script/system_check
```diff
@@ -0,0 +1,31 @@
...
+}
+
+# Use this script to add any processes required to boot your application.
+# A `print_error` function is provided as a helpful way to display a custom
+# error message, and exit the script when a process fails, or is missing.
+#
+# A few examples might be:
+#   * `lucky db.verify_connection` to test postgres can be connected
+#   * Checking for elasticsearch, redis, or any other process is booted first
+#   * Booting additional processes for things like mail, background jobs, etc...
```
paulcsmith: Booting additional processes should probably go in the Procfile.dev. Maybe that can be noted here?

src/web_app_skeleton/script/system_check
```diff
@@ -0,0 +1,31 @@
...
+    printf "${BOLD_RED_COLOR}There is a problem with your system setup:\n\n   ${BOLD_RED_COLOR}$1 \n\n"
+    exit 1
+}
+
+# Use this script to add any processes required to boot your application.
+# A `print_error` function is provided as a helpful way to display a custom
+# error message, and exit the script when a process fails, or is missing.
+#
+# A few examples might be:
+#   * `lucky db.verify_connection` to test postgres can be connected
```
paulcsmith: This is a great example. I kind of think we should run this by default. What do you think?

src/web_app_skeleton/script/system_check
paulcsmith: I'd put this as a comment otherwise I don't think people will know what it does. I definitely don't heh

src/web_app_skeleton/Procfile.dev.ecr
paulcsmith: I *think* you can script/system_check without the `./`

paulcsmith: Also thoughts on `check_system` so it is more of a verb? Like "Hey computer, check my system". Kind of like if we had system in `setup`, setup_system` would read better than `system_setup` IMO

What do you think

paulcsmith:
I'm very excited about having a convention for where to put these checks. In almost every codebase I've been on there are notes in the README (if I'm lucky) and checks in setup (again, if I'm lucky haha).

Having a clear place will really help with this, especially if we have some nice default ones like crystal, yarn, postgres, etc. checks

src/web_app_skeleton/Procfile.dev.ecr
jwoertink: So `check_system: script/check_system`? or `system_check: script/check_system` like the first part is the noun and the second part is the verb? @edwardloveall ^?

paulcsmith: Yeah, I was thinking `check_system: script/check_system` but would love to hear what Edward thinks too :)

edwardloveall: I think I'm partial to check system? but honestly this might be bikeshedding too much. I think maybe we're struggling because they're both clear. The only way they could be more clear is if you added an article like `check_the_system` or a subject like `system_check_yourself_before_you_wreck_yourself`. I have no idea if that's correct and I'm mostly making this up. I think you could keep it as-is and we can switch it if it becomes "wrong" later.

jwoertink: Oh, I think you are correct. You definitely have to check yourself because wrecking yourself is a very real thing. 

I'll roll with `check_system`. After saying a few times, I like it better.

jwoertink:
Just put this in to my project, and it's working out great

```
[08:50AM] app (develop)$ lucky dev
system       | Tmux socket name: overmind-vip-YKZUn_NBtrxnD_xJW4ZavL
system       | Tmux session ID: vip
system       | Listening at /Users/jeremywoertink/Sites/app/.overmind.sock
web          | Started with pid 3093...
check_system | Started with pid 3092...
check_system | Could not connect to Redis at 127.0.0.1:6379: Connection refused
check_system | There is a problem with your system setup:
check_system |
check_system |    Redis is not currently running. Be sure to boot it first.
check_system |
web          | Beginning to watch your project
web          | Compiling...
check_system | Exited
web          | Interrupting...
web          | Exited
[08:50AM] app (develop)$
[08:50AM] app (develop)$
[08:50AM] app (develop)$ brew services start redis
==> Successfully started `redis` (label: homebrew.mxcl.redis)
[08:50AM] app (develop)$
[08:50AM] app (develop)$
[08:50AM] app (develop)$ lucky dev
system       | Tmux socket name: overmind-vip-QcVN00O4DPDY2h38uSsunr
system       | Tmux session ID: app
system       | Listening at /Users/jeremywoertink/Sites/app/.overmind.sock
check_system | Started with pid 3397...
web          | Started with pid 3398...
check_system | PONG
web          | Beginning to watch your project
web          | Compiling...
web          | Done compiling
web          | Listening on http://0.0.0.0:5100
web          |
web          |
web          |    🎉 App running at http://0.0.0.0:5100
web          |
web          |
```

The one I'm using is:

```bash
redis-cli ping
if [ $? -ne 0 ]  # Check the exit_code of the previous task. If not equal to 0, then print_error
then
  print_error "Redis is not currently running. Be sure to boot it first."
fi
```

wontruefree:
I really like this idea

jwoertink:
Ok, after bashing my head against this code, I was able to get some other stuff working as well specific for my apps. I wonder if this should include some bash guides, or helpful tips? Maybe even a few helper methods to check some things?

src/web_app_skeleton/script/check_system
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+BOLD_RED_COLOR="\e[1m\e[31m"
+
+# Prints error and exit.
+# example:
+#   print_error "Redis is not running. Run it with some_command"
+print_error() {
+    printf "${BOLD_RED_COLOR}There is a problem with your system setup:\n\n   ${BOLD_RED_COLOR}$1 \n\n"
```
paulcsmith: I think passing in the text without this prefix is probably enough. Like "Redis isn't running" or "postgres isn't running. Please start it." is sufficient. Thoughts?

src/web_app_skeleton/script/check_system
```diff
@@ -0,0 +1,32 @@
...
+    exit 1
+}
+
+# Use this script to add any processes required to boot your application.
```
paulcsmith:
```suggestion
# Use this script to check the system for required tools and process that your app needs.
```

Since this could also be used to do generic checks like crystal version, yarn installed, etc.

src/web_app_skeleton/script/check_system
```diff
@@ -0,0 +1,32 @@
...
+# A `print_error` function is provided as a helpful way to display a custom
+# error message, and exit the script when a process fails, or is missing.
+#
+# A few examples might be:
+#   * `lucky db.verify_connection` to test postgres can be connected
+#   * Checking for elasticsearch, redis, or any other process is booted first
```
paulcsmith:
```suggestion
#   * Checking that elasticsearch, redis, or postgres is booted
```

src/web_app_skeleton/script/check_system
```diff
@@ -0,0 +1,32 @@
...
+#   * Booting additional processes for things like mail, background jobs, etc...
+#     should go in your Procfile.dev.
+
+lucky db.verify_connection
+if [ $? -ne 0 ]  # Check the exit_code of the previous task. If not equal to 0, then print_error
+then
+  print_error "Postgres does not seem to be running. Be sure you can connect."
```
paulcsmith:
```suggestion
  print_error "Postgres does not seem to be running or the credentials are wrong. Check config/database.cr and be sure you can connect."
```

src/web_app_skeleton/script/check_system
jwoertink: You mean get rid of this top line "There is a problem...", and just leave the bottom part?

Screen Shot 2020-03-27 at 1 14 32 PM

paulcsmith: Never mind, I like that a lot :) I didn't realize it had some newlines! That looks super nice 👍

paulcsmith: I'm bad at reading shell scripts :)

jwoertink:
Ok, after this is merged in, I'll open up an issue about writing some docs on it.

I was thinking about the setup stuff coming over to this, but had a concern. I originally saw this as purely a way to check and make sure things that need to be running for you to work on the app are actually running. I'm wondering if it would slow down the boot time if it has to check these on every boot:
* is process runner installed? if this is called from lucky dev, then obviously it is.
* is yarn installed?
* is postgres installed?
* is postgres running?
* is crystal installed?
* maybe is crystal the right version?
* whatever else....

To me, these make sense to check when you setup your app, but once you've done that, you don't need to check them again. Unless you randomly uninstalled yarn or pg? but that feels edge casey... 

Let me know if I'm just over thinking this 😅  I guess we could throw them in and try it out, if it gets clunky, we just rip it out. Pre 1.0 is the time to experiment 😆

paulcsmith:
@jwoertink I think that those checks are super fast typically, also system_check will run in a separate process so will not block yarn or crystal compilation. The other part of this is what if you have a new check in setup because someone added elasticsearch to the setup checks. If you already had the app setup you'd start with `lucky dev` and unless they also checked in check_system they would not know about it.

The other part is that sometimes people just don't run the setup script 🤣 so having this would be an extra layer of helpful checks.

Does that make sense?

jwoertink:
ah! Ok, `if you have a new check in setup because someone added elasticsearch` is what sold me. I've had that happen a few times. Gone to other devs like "this won't boot", then get "oh yeah, you have to `brew install whatever` first...". 😐 

Ok, I'll add that in... Also, side note, you've called it `system_check` twice now which makes me wonder if my original naming was better 😛

paulcsmith:
Yeah I noticed that. I think you were right. I intuitively call it system_check and seems you did too so we can switch it back. My bad! 🤣

jwoertink:
Oh, here's a question. In the setup script, we create the DB, then check to see we can connect before migrating. We are also saying postgres has to be connected before booting. It's a whole chicken an egg deal now when running a setup because it'll be like:

When running setup
* check connection.. fail
* create db
* check connection... ok
* migrate

When running lucky dev, it won't matter. Should the setup script just catch that first error and skip past it?

jwoertink:
maybe I can just pass in a flag on whether or not it should be checked. Add it in the Procfile, but tell it no for setup

paulcsmith:
Ok thinking we don’t do the verify Db in the system check and only do it in setup so the script doesn’t get too complex. Kind of a bummer to not have that but I think it’ll be ok since we have good errors when it can’t connect

src/web_app_skeleton/script/system_check.ecr
```diff
@@ -0,0 +1,46 @@
...
+
+# Only if this isn't CI
+if [ -z "$CI" ]; then
+  lucky ensure_process_runner_installed
+fi
+
+if ! command -v createdb > /dev/null; then
+  MSG="Please install the postgres CLI tools, then try again."
+  if [[ "$OSTYPE" == "darwin"* ]]; then
+    MSG="$MSG\nIf you're using Postgres.app, see https://postgresapp.com/documentation/cli-tools.html."
```
paulcsmith:
```suggestion
    MSG="$MSG\n  If you're using Postgres.app, see https://postgresapp.com/documentation/cli-tools.html."
```

I think this needs 2 extrac spaces since `print_error` indents a bit`

src/web_app_skeleton/script/system_check.ecr
```diff
@@ -0,0 +1,46 @@
...
+  fi
+  MSG="$MSG\nSee https://www.postgresql.org/docs/current/tutorial-install.html for install instructions."
+
+  print_error "$MSG\n"
```
paulcsmith: Actually after thinking about this maybe we `source script/text_helpers` and use `indent` in the `print_error` function. That way each new line is auto indented:

```
printf "${BOLD_RED_COLOR}There is a problem with your system setup:${BOLD_RED_COLOR}\n\n"
indent $1
```

Not sure if that would work, but would be cool if it did

src/web_app_skeleton/script/helpers/function_helpers
```diff
@@ -0,0 +1,67 @@
...
+  fi
+}
+
+# Returns true if the OS is linux based
+# example:
+#   if is_linux; then
+#     echo "do linux stuff"
+#   fi
+is_linux() {
+  if [[ "$OSTYPE" == "linux"* ]]; then
```
jwoertink: It seems most people I asked get back `linux-gnu` on arch, debian, or other flavors. I didn't actually test this though, but I assume it works.

src/web_app_skeleton/script/setup.ecr
```diff
@@ -4,69 +4,44 @@
...
-}
 
-<%- if browser? -%>
-if ! command -v yarn > /dev/null; then
-  printf 'Yarn is not installed.\n'
-  printf 'See https://yarnpkg.com/lang/en/docs/install/ for install instructions.\n'
-  exit 1
-fi
+notice "Running System Check"
+INIT_SETUP=0 ./script/system_check
```
jwoertink: This tells the system_check to not return the `$SHELL`

src/web_app_skeleton/script/system_check.ecr
```diff
@@ -0,0 +1,53 @@
...
+fi
+
+
+## CUSTOM PRE-BOOT CHECKS ##
+# example:
+# if command_not_running "redis-cli ping"; then
+#   print_error "Redis is not running."
+# fi
+
+
```
jwoertink: My thought was having a very clear place to put your own custom stuff. This file is already really busy, so if you're new to bash, then it might not be apparent where you'd do stuff, or what you'd do.

src/web_app_skeleton/script/system_check.ecr
```diff
@@ -0,0 +1,53 @@
...
+## END ##
+
+
+
+# Skip this during initial setup
+if [ -z "$INIT_SETUP" ]; then
+  # Return $SHELL if no error is raised to allow the process runner to continue.
+  $SHELL
```
jwoertink: This needs to be returned when it's called from the `Procfile.dev`, but should be skipped when called from the `setup` script.

jwoertink:
Ok, I added a bunch of goodies here. I threw this in to a fresh 0.19 app and it worked great. If y'all wanna give it a once over that'd be cool, but I think it's ready to go now.

src/web_app_skeleton/Procfile.dev.ecr
```diff
@@ -1,3 +1,4 @@
+system_check: script/system_check
```
paulcsmith: I think it’d be cool to have the system check never return SHELL so you can use it from other scripts or just run it on its own. Instead I think you could modify this line to `script/system_check && $SHELL` since shell is only necessary in the Procfile. That way the system check doesn’t have any conditional logic

src/web_app_skeleton/script/helpers/function_helpers
```diff
@@ -0,0 +1,67 @@
...
+# functions you may need for your app
+
+
+# Returns true if the command $1 is not found
+# example:
+#   if command_not_found "yarn"; then
+#     echo "no yarn"
+#   fi
+command_not_found() {
+  ! command -v $1 > /dev/null
```
paulcsmith: I think it would be more flexible if  this was just “command” and the example would be “yarn -v”. Because some commands need “—version”

src/web_app_skeleton/script/helpers/function_helpers
```diff
@@ -0,0 +1,67 @@
...
+    true
+  else
+    false
+  fi
+}
+
+# Prints error and exit.
+# example:
+#   print_error "Redis is not running. Run it with some_command"
+print_error() {
```
paulcsmith: I think this makes sense in the text helpers script but up to you. I don’t feel strongly about it

src/web_app_skeleton/script/helpers/function_helpers
jwoertink: I don't think that's what that means. I'm under the impression that the `-v` flag is on `command`, and not `$1`. I'm having trouble finding the exact documentation on it. [according to this thread](https://stackoverflow.com/questions/592620/how-can-i-check-if-a-program-exists-from-a-bash-script) some say the same, and some say `-v` doesn't work in some shells (like fish). But we're only really gearing toward bash and zsh compliant.

src/web_app_skeleton/script/helpers/function_helpers
jwoertink: I thought that at first, but because it does an exit, and has specific text in it, it felt more like a function to me

src/web_app_skeleton/script/helpers/function_helpers
jwoertink: Oh, actually, this confirms it:

```
[14:50PM] myapp (master)$ command -v yarn
/usr/local/bin/yarn
[14:51PM] myapp (master)$ command yarn
yarn install v1.19.1
[1/4] 🔍  Resolving packages...
success Already up-to-date.
✨  Done in 0.46s.
[14:51PM] myapp (master)$ command --version yarn
zsh: command not found: --version
```

The `command -v` is basically replacing `which` which seems to be not as posix compliant? 🤔  so I'm gonna keep it like this.

src/web_app_skeleton/Procfile.dev.ecr
jwoertink: I like that better.

PR #483, opened 2020-03-23
See if there are errors on master

paulcsmith:
I think the changes I made to Lucky::Params somehow messed things up and thankfully the full integration suite in this repo caught them. 

I first saw this in https://github.com/luckyframework/lucky_cli/pull/481, which I don't think has anything to do with the compile message. So testing without those changes to be sure

### DIFF
--- a/src/lucky.cr
+++ b/src/lucky.cr
@@ -5,6 +5,7 @@ require "./dev"
 require "./ensure_process_runner_installed"
 
 include LuckyCli::TextHelpers
+# Changed
 
 args = ARGV.join(" ")
 tasks_file = ENV.fetch("LUCKY_TASKS_FILE", "./tasks.cr")


paulcsmith:
Yep the params thing broke it. Will investigate. I think I know what change broke it, but I'm not sure why 🤔

wontruefree:
Is this on release 0.19.0?

paulcsmith:
@wontruefree No it is in Lucky master so 0.19 is good :)

paulcsmith:
Here is the PR and the specific line that I think caused the issue: https://github.com/luckyframework/lucky/pull/1053/files#diff-d633a5eeaf04a8d42861bcc3614b84bfR481

paulcsmith:
Fixed: https://github.com/luckyframework/lucky/pull/1061

PR #485, opened 2020-03-24
Add default with comments for pipe log

paulcsmith:
Part of: https://github.com/luckyframework/lucky/pull/1062

### DIFF
--- a/src/web_app_skeleton/config/logger.cr
+++ b/src/web_app_skeleton/config/logger.cr
@@ -33,6 +33,20 @@ Lucky.configure do |settings|
   settings.logger = logger
 end
 
+Lucky::Action.configure do |settings|
+  # By default Lucky only logs when before/after pipes halt by redirecting, or
+  # rendering a response.
+  #
+  # Set 'pipe_log_level' to a 'Logger::Severity' to log whenever any pipe runs.
+  #
+  # For example:
+  #
+  #    settings.pipe_log_level = Logger::Severity::DEBUG
+  #
+  # Full list of log levels: https://crystal-lang.org/api/Logger/Severity.html#enum-members
+  settings.pipe_log_level = nil
+end
+
 Lucky::LogHandler.configure do |settings|
   # Skip logging static assets in development
   if Lucky::Env.development?


src/web_app_skeleton/config/logger.cr
```diff
@@ -31,6 +31,15 @@ logger =
...
   settings.logger = logger
+  # By default Lucky only logs when before/after pipes halt (by redirecting, or
+  # rendering a response).
+  #
+  # Set the log level to log to log every pipe that runs.
+  #
+  # For example:
+  #
+  #    settings.action_pipe_log_level = Logger::Severity::DEBUG
+  settings.action_pipe_log_level = nil
```
paulcsmith: Will fail until https://github.com/luckyframework/lucky/pull/1062 is merged.

paulcsmith:
@edwardloveall @jwoertink this will fail until the other PR is merged, but could you take a look at the comment and see if it makes sense?

src/web_app_skeleton/config/logger.cr
edwardloveall: Can we also list the other values that `action_pipe_log_level` can be?

src/web_app_skeleton/config/logger.cr
```diff
@@ -31,6 +31,15 @@ logger =
 
 Lucky.configure do |settings|
   settings.logger = logger
+  # By default Lucky only logs when before/after pipes halt (by redirecting, or
+  # rendering a response).
+  #
+  # Set the log level to log to log every pipe that runs.
```
jwoertink:
```suggestion
  # Set the log level to log every pipe that runs.
```

paulcsmith: Oops :)

src/web_app_skeleton/config/logger.cr
paulcsmith: There are a bunch but I can link to the Logger docs that show all the possible enums!

paulcsmith:
@jwoertink if it is nil it means they will only log if they are halted by returning a redirect, json, html, etc.

Maybe we should add a resuce as well that logs which pipes have an error, but I'm not sure since that could get hairy and could be caught from the stacktrace

paulcsmith:
I think I addressed everything. Still waiting on the other PR to be green and then I can re-run this on CI

PR #497, opened 2020-04-05
Use same path for New/Create

paulcsmith:
Closes https://github.com/luckyframework/lucky/issues/1003


src/browser_authentication_app_skeleton/src/actions/sign_ins/create.cr
```diff
@@ -1,7 +1,7 @@
 class SignIns::Create < BrowserAction
   include Auth::RedirectSignedInUsers
 
-  route do
+  post "/sign_in" do
```
jwoertink, line 4: Nice! Yeah, we should definitely make sure all of these are gone. Will be easier to rip this out later if that's what we end up doing.

PR #499, opened 2020-04-05
Fix Errno issue with Crystal 0.34.0 and release v0.20.0

paulcsmith:
Errno was renamed in 0.34.x https://github.com/crystal-lang/crystal/pull/8885

Instead we should just ignore non existent file. That way it works in
0.33 and 0.34

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -4,11 +4,11 @@ version: 0.19.0
 authors:
   - Paul Smith <paulcsmith0218@gmail.com>
 
-crystal: 0.33.0
+crystal: 0.34.0
 
 license: MIT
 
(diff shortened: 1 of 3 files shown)


paulcsmith:
Issue with shards not resolving since we can't override the dependency. I don't want to go create separate branches for every dependency so they point to lucky master. So hoping https://github.com/crystal-lang/shards/issues/299 gets added. 

If not added by 0.34.0 (which seems quite likely it won't be since 0.34 is pretty close) then I'll wait to get a version of lucky locked down for 0.34.0, then update this to point to that branch and should hopefully be able to test it out

PR #500, opened 2020-04-06
[DO NOT MERGE] Crystal 0.34 compat and update version to 0.19.2

paulcsmith:
Will be tagged and closed once CI is green


src/web_app_skeleton/.crystal-version
```diff
@@ -1 +1 @@
-0.32.1
+0.33.0
```
jwoertink, line 1: This should be 0.34.0, right?

paulcsmith: Oh yeah good catch

paulcsmith:
I need to update LuckyFlow as well, and then this should work 🤞

paulcsmith:
I don't know why this is failing. Going to close and open a new PR just to make sure Travis isn't caching stuff or doing something strange

PR #508, opened 2020-04-25
Memoize current_user in actions

paulcsmith:
Still need to do this in ApiAction too

### DIFF
--- a/src/browser_app_skeleton/src/actions/browser_action.cr.ecr
+++ b/src/browser_app_skeleton/src/actions/browser_action.cr.ecr
@@ -25,7 +25,7 @@ abstract class BrowserAction < Lucky::Action
   expose current_user
 
   # This method tells Authentic how to find the current user
-  private def find_current_user(id) : User?
+  memoize def find_current_user(id) : User?
     UserQuery.new.id(id).first?
   end
   <%- end -%>
--- a/src/browser_app_skeleton/src/actions/browser_action.cr.ecr
+++ b/src/browser_app_skeleton/src/actions/browser_action.cr.ecr
@@ -25,7 +25,8 @@ abstract class BrowserAction < Lucky::Action
   expose current_user
 
   # This method tells Authentic how to find the current user
-  memoize def find_current_user(id) : User?
+  # The 'memoize' macro makes sure only one query is issued to find the user
+  private memoize def find_current_user(id) : User?
     UserQuery.new.id(id).first?
   end
   <%- end -%>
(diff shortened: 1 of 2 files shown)


paulcsmith:
Memoize does not work with a "Nil". Opened an issue on Lucky to support this use case

src/browser_app_skeleton/src/actions/browser_action.cr.ecr
```diff
@@ -25,7 +25,8 @@ abstract class BrowserAction < Lucky::Action
   expose current_user
 
   # This method tells Authentic how to find the current user
-  private def find_current_user(id) : User?
+  # The 'memoize' macro makes sure only one query is issued to find the user
+  private memoize def find_current_user(id) : User?
```
jwoertink, line 29: You can just set this type to `User`. If it's nil, it just won't memoize. I saw you opened the issue already for that, but until we have that in, this could be a small win.

paulcsmith: I'm not sure I understand. The User can be nil so I can't change the return type to not have nil.

jwoertink: The `memoize` right now doesn't allow you to specify a union like `User?` https://github.com/luckyframework/lucky/blob/master/src/lucky/memoizable.cr#L21 so this should throw a compile time error.

paulcsmith: Yeah this does throw an error so I opened the issue to add support. I’m not sure how we’d get a quick win by setting it to User because that also won’t compile. I am quite tired so maybe just being dense haha

jwoertink: Now that memoize supports taking args and having a nil value, I think this is fine. 😄

wontruefree:
when there is a query cache will this make sense?

paulcsmith:
@wontruefree It still might be good so that it doesn't log a cache hit, but the performance gain would be very minimal once the query cache is shipped

wontruefree:
I am curious because memorization like this will cause more memory usage for minimal performance with query caching.  This might be worth revising when query cache is added.

paulcsmith:
Good idea! We can revisit once query caching is added 👍

jwoertink:
Just dropping some notes here. Currently neither of these updates would compile. Bug filed here https://github.com/luckyframework/lucky/issues/1396

jwoertink (approved):
Well, it took a few years, but this change actually works now :laughing:

PR #516, opened 2020-06-10
Pin jwt on 1.4.1 untill Crystal 0.35 is supported

richarddewit:
Version [1.4.2 of crystal-community/jwt](https://github.com/crystal-community/jwt/releases/tag/v1.4.2) made a breaking change of requiring a new method only available from Crystal v0.35.0, which gives an error when running `lucky dev`.

[Source of change mentioned](https://github.com/crystal-community/jwt/commit/29a3b1cb910ae12f9728f0598124ebe841eb71b8)

### DIFF
--- a/src/generators/web.cr
+++ b/src/generators/web.cr
@@ -177,7 +177,7 @@ class LuckyCli::Generators::Web
       append_text to: "shard.yml", text: <<-DEPS_LIST
         jwt:
           github: crystal-community/jwt
-          version: ~> 1.4.0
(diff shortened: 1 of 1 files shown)


paulcsmith:
Releasing v0.22.0 with support for Crystal 0.35 so this won't be necessary! We'll have upgrade notes up soon, but the gist is to change shard.yml to use authentic 0.6.0, jwt 1.4.2, and lucky 0.22.0

PR #517, opened 2020-06-11
Fix compatibility issues with babel and Node 13

paulcsmith:


src/browser_app_skeleton/package.json
```diff
@@ -4,7 +4,8 @@
   "dependencies": {
     "@rails/ujs": "^6.0.0",
     "normalize-scss": "^7.0.1",
-    "turbolinks": "^5.2.0"
+    "turbolinks": "^5.2.0",
+    "@babel/compat-data": "^7.9.0"
```
jwoertink, line 8: should this be a devDependency?

paulcsmith: I think that I added it in devDependency and it did not work. But I will try again!

paulcsmith: Ok I have weird node paths for my local environment so I can get GitHub running locally. I have to push to heroku to test. I think I'll merge this as-is since I am 99% sure it didn't work as a devDependency (it was needed to build with `yarn prod`). So I'll merge and we can reevaluate later. I want to try to get Lucky out and working pretty soon

PR #518, opened 2020-06-12
Remove FlashHandler and SessionHandler.

wout:
The final part of the session bug fix. :)


paulcsmith:
Travis has been acting up lately. Gonna go ahead and merge this. We need to just switch to GH actions since they're much easier to use and set up

PR #519, opened 2020-06-15
Use latest teeplate and Crystal 0.35.0

paulcsmith:


paulcsmith:
There is an issue with LuckyFlow but this will be resolved with the new version of LuckyFlow that auto-installs the right chromedriver

PR #521, opened 2020-06-17
Adds in new lucky db.setup task to replace calling create and migrate

### LINKED ISSUE #513: Use new db.setup task in generated apps
https://github.com/luckyframework/avram/pull/361 added in a new `db.setup` task which does a `db.create` and `db.migrate` together. By adding this in to our generated apps it should speed things up a teensy bit during initial setup.

jwoertink:
Fixes #513

Instead of calling create and migrate separately, we can now just call `db.setup` which does both in a single process. This will speed up setup just slightly. Also removed the `db.verify_connection` because you can't call it before the db exists otherwise connection fails. And if the `db.setup` works, then the `verify_connection` has already been met. That makes this irrelevant. 

For a future idea, maybe we can pass an arg to verify_connection that just ensures you have a connection to the PG engine itself. For example, you can connect to AWS RDS with these credentials, so you have the ability to call create or migrate if you want..

### DIFF
--- a/src/web_app_skeleton/script/setup.ecr
+++ b/src/web_app_skeleton/script/setup.ecr
@@ -30,14 +30,8 @@ if [ ! -f ".env" ]; then
   print_done
 fi
 
-notice "Creating the database"
-lucky db.create | indent
-
-notice "Verifying postgres connection"
-lucky db.verify_connection | indent
-
-notice "Migrating the database"
-lucky db.migrate | indent
+notice "Creating and Migrating the database"
+lucky db.setup | indent
 
 notice "Seeding the database with required and sample records"
 lucky db.create_required_seeds | indent


src/web_app_skeleton/script/setup.ecr
```diff
@@ -30,14 +30,8 @@ if [ ! -f ".env" ]; then
...
-notice "Creating the database"
-lucky db.create | indent
-
-notice "Verifying postgres connection"
-lucky db.verify_connection | indent
-
-notice "Migrating the database"
-lucky db.migrate | indent
+notice "Creating and Migrating the database"
+lucky db.setup | indent
```
paulcsmith, line 34: Does `lucky db.setup` raise a nice error like `db.verify_connection` does? I think it makes sense to make sure all the db tasks return a nice error if it can't connect. If we don't have a nice error for `db.setup` I think we should keep this as-is and switch to db.setup when the errors are nice.

But maybe they already are nice, in which case I think this LGTM :)

jwoertink: It will try to create the DB first, and if it can't connect, it throws

```
Creating the database failed. It looks like Postgres is not running.
      Message from Postgres:
        #{message}
      Try this...
        ▸ Make sure Postgres is running
        ▸ Check your database configuration settings
```

I do like the error message `verify_connection` gives, but if the DB doesn't exist yet, then it fails, so you have to call `create_db` first anyway.

I could update the `verify_connection` so the DB name is optional. Then if you pass it, it would check if you can connect to that instance, and if it's not there, then it just checks that you can connect to that engine... (not sure how that works yet).

jwoertink:
Ok, I didn't think this whole deal out too well. Now that `lucky db.create` will throw a nice connection error if it can't connect, I realized that the `db.setup` never calls `db.create`... it calls `Runner.db_create`... Then I realized that that method already handles doing it's own connection error message stuff which seriously needs to be cleaned up. I think I'm just gonna skip this for now, and this can be added back in during our "spring cleaning".

PR #525, opened 2020-06-18
Some updates for Crystal 0.35 and latest release

jwoertink:
Due to some funky stuff with the shards, I have to either tell the generated app to be on 0.22, or cut a new release of Authentic that tells Lucky to be on master branch. Running specs locally all fail with this update... but my last PR wouldn't even run the specs, so I have no idea if anything is actually broken, or just pretending to be broken...


paulcsmith:
What I'd do is make a branch for all of the shards that have big updates, like `jw/v0.x.0`. Then point all the shards to the branches.

Then once everything is green, you'd tag those branches with the version number, then update all the shards to use the new version instead of the branch. This is confusing and takes a lot of repitious work, but there is no other reliable way to do it until shards gets an `override` option :(

jwoertink:
Ok, the current failure is due to chromedriver throwing:

> bind() failed: Cannot assign requested address (99)

which could either mean there's a random chromedriver process not dying when it should, or something like it's trying to connect on ipv6, but that's not enabled or something weird.... In any case, the current flow is pointed to an old version. I'm going to merge this in, and then start working on pointing everything to custom new branches and see how all that goes.

PR #526, opened 2020-06-19
remove travis and add Gh actions to check format and run specs

jwoertink:
Not sure if I have to do anything else, but let's see how this goes.


jwoertink:
Ok, well that's a good sign. I get those same errors locally... So now to figure out why

jwoertink:
Actually, I'm going to merge this in since this PR is technically now implemented! 🎉

PR #534, opened 2020-07-14
Renamed AppClient to ApiClient.

### LINKED ISSUE #492: AppClient is always sending json Content-Type
When we generate a project, we add this `AppClient`, but it always defaults the Content-Type to json. If you're wanting to test just normal web actions, then you always have to override this. We should just add the catch `<% if api_only? %>`, then it's a safe bet we want this. If you're doing an XML API (or whatever), then you can override this on your own 😛 

https://github.com/luckyframework/lucky_cli/blob/0b2f8bf859dd693b09e3c9ef658cdb1ac6c0c697/src/web_app_skeleton/spec/support/app_client.cr.ecr#L4

jwoertink:
Fixes #492


src/browser_app_skeleton/spec/support/browser_client.cr.ecr
```diff
@@ -0,0 +1,12 @@
+class BrowserClient < Lucky::BaseHTTPClient
```
paulcsmith, line 1: Will this work correctly for people when they try to set params? I think since the content_type is set to text/html Lucky will not be able to parse params correctly. But I'm not sure. Same with the header stuff. I think it needs a session.

Unless `BrowserClient` works with params and auth as expected I think this should be removed and added later when we can make it work with sessions and params.

One way to handle params is to not set the content type and instead use application/json with json params. Otherwise we'll need to change the BaseHTTPClient (I think) to send form encoded params somehow

jwoertink: Oh, you're saying when doing `BrowserClient.post()` because the params might be expecting the formenc type or whatever it is, this would clash... 🤔  good point. Ok, so for this PR then, we just rename `AppClient` to `ApiClient`, and in another PR add `BrowserClient` which can handle doing sessions and params and all that....

paulcsmith: Yeah exactly! 👍

PR #535, opened 2020-07-18
added disable_cookies as default for api actions

confact:
Added disable cookies as default for API actions

### DIFF
--- a/src/web_app_skeleton/src/actions/api_action.cr.ecr
+++ b/src/web_app_skeleton/src/actions/api_action.cr.ecr
@@ -1,6 +1,7 @@
 # Include modules and add methods that are for all API requests
 abstract class ApiAction < Lucky::Action
-  # This disable setting cookies and session header. Cookie header will still be sent.
+  # This disable setting cookies and session header.
+  # Remove it to set set-cookie header to use it in api actions
   disable_cookies
   accepted_formats [:json]
   <%- if generate_auth? -%>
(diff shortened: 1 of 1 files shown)


src/web_app_skeleton/src/actions/api_action.cr.ecr
```diff
@@ -1,5 +1,6 @@
 # Include modules and add methods that are for all API requests
 abstract class ApiAction < Lucky::Action
+  disable_cookies
```
jwoertink, line 3: Can we add a little comment on what this is removing? It removes sessions too, doesn't it?

confact:
@jwoertink Now API action's current user is using JWT as default and should not need cookies and sessions. If they want to use API action with cookies and sessions they would need to use the module as in the Browser action anyway.

Like in API Action we include:
```
include Api::Auth::Helpers
include Api::Auth::RequireAuthToken
```
But in browser action we include:
```
include Auth::RequireSignIn
```

I can add comments for it though to know why it is there and can be removed if they want to use session and cookies (they would need to change `Auth::Helpers` and `RequireAuthToken` as well)

src/web_app_skeleton/src/actions/api_action.cr.ecr
confact: Added comment @jwoertink - feel free to change it to something better :)

jwoertink:
Ah, that makes sense. 👍

src/web_app_skeleton/src/actions/api_action.cr.ecr
paulcsmith: Thanks for catching this and adding it. It makes sense to disable cookies by default IMO! Is this right that "Cookie header will still be sent"? I thought the header wouldn't be sent if cookies are disabled?

confact: @paulcsmith you are right. I must have been confused of the request headers. I change the comment now.

Feel free to merge it if it looks good :)

PR #538, opened 2020-08-05
Add Ameba checks to CI and fix non-compliant code

stephendolan:
This pull request seeks to do a few things:

- Adds a checker to the `script/test` script to ensure that `crystal tool format` passes
- Adds a checker to the CLI code itself in both `script/test` and the CLI pipeline to ensure that the Ameba linter passes
- Brings the CLI code itself into a state where it passes the Ameba linter
- Brings the generated application code into a state where any `lucky init` app should pass the Ameba linter

I was having trouble running these tests locally (`crystal spec` was failing a single integration, but running the integration on its own passed), so I'd like to see the GitHub action results on this PR before changing anything else.

As always, open to feedback on a change in direction for any of this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ on:
     branches: "*"
 
 jobs:
-  lint:
+  check_format:
     runs-on: ubuntu-latest
     container:
       image: crystallang/crystal:0.35.0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Format
         run: crystal tool format --check
       - name: Crystal Ameba Linter
-        uses: crystal-ameba/github-action@v0.2.5
+        uses: crystal-ameba/github-action@v0.2.6
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,12 +2,12 @@ name: Lucky CLI CI
 
 on:
   push:
-    branches: [ master ]
+    branches: [master]
   pull_request:
     branches: "*"
 
 jobs:
-  check_format:
+  lint:
     runs-on: ubuntu-latest
     container:
       image: crystallang/crystal:0.35.0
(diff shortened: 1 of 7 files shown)


stephendolan:
I'll look into and remedy the GitHub Action issues once I receive some general feedback on the usefulness of this PR to the project!

src/api_authentication_app_skeleton/src/models/user_token.cr
```diff
@@ -13,8 +13,10 @@ class UserToken
     JWT.encode(payload, Lucky::Server.settings.secret_key_base, ALGORITHM)
   end
 
+  # Both the payload and header are provided from the JWT decode method
+  # Should you require the header, modify the `_` variable below to something like `header`
   def self.decode_user_id(token : String) : Int64?
-    payload, _header = JWT.decode(token, Lucky::Server.settings.secret_key_base, ALGORITHM)
+    payload, _ = JWT.decode(token, Lucky::Server.settings.secret_key_base, ALGORITHM)
```
paulcsmith, line 19: I'm pretty against `_` because it is so unclear it needs a comment. I wonder if we can disable this rule in Ameba? I'd say for me this is a dealbreaker. Probably quite pedantic, but I think it is helpful to be able to label unused variables so you don't need comments.

stephendolan: Totally agreed, @paulcsmith. The only trouble is that unless we *also* disabled this in a default-included `.ameba.yml` file, every newly-generated Lucky app will still produce an Ameba error by default.

@jwoertink and I talked about this a bit in Gitter, and maybe it's worth leaving this PR on hold until I open and resolve an issue with the Ameba team directly.

IMO, there's no reason that you *wouldn't* want to show what an unused variable actually is in case you need it in the future.

stephendolan: The newest version of Ameba is cool with this, so it's been reverted in this PR!

.github/workflows/ci.yml
```diff
@@ -17,6 +17,9 @@ jobs:
         run: shards install
       - name: Format
         run: crystal tool format --check
+      - name: Crystal Ameba Linter
+        run: crystal bin/ameba.cr
```
jwoertink, line 21: oh, I wonder if this should just be https://github.com/luckyframework/lucky/blob/master/.github/workflows/ci.yml#L21 ?

stephendolan: I think either one will work, but I'm all for keeping things consistent among the org's repos. Changed!

stephendolan:
And we're green! @paulcsmith, this has been resolved with the help of the lovely Ameba team such that any `lucky init` app will now pass Ameba's checks, and this repo itself will have Ameba checks against all commits moving forward.

PR #543, opened 2020-08-27
Added a catch for reserved project names

### LINKED ISSUE #540: Creating a project named as 'app' makes project broken.
If you create a project named as `app`, lucky cli will overwrite the original file `app.cr` in [web_app_skelton](https://github.com/luckyframework/lucky_cli/tree/master/src/web_app_skeleton/src).
This make project completely broken.

jwoertink:
Fixes #540

There's a small chance that when someone is trying Lucky for the first time, they may just generate an app called "app" just to get through the wizard. Doing this would overwrite the `src/app.cr` file which causes some funky stuff. In this PR I just added all of the pre-existing file names to avoid any of that.

### DIFF
--- a/src/lucky_cli/project_name.cr
+++ b/src/lucky_cli/project_name.cr
@@ -20,20 +20,18 @@ class LuckyCli::ProjectName
       <<-TEXT
       Project name can't be blank
       TEXT
+    elsif RESERVED_PROJECT_NAMES.includes?(sanitized_name)
+      <<-TEXT
+      Projects cannot be named #{RESERVED_PROJECT_NAMES.join(", ")}.
+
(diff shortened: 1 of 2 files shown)


src/lucky_cli/project_name.cr
```diff
@@ -19,11 +21,19 @@ class LuckyCli::ProjectName
       Project name can't be blank
       TEXT
     else
-      <<-TEXT
-      Project name should only contain lowercase letters, numbers, underscores, and dashes.
+      if RESERVED_PROJECT_NAMES.includes?(sanitized_name)
```
paulcsmith, line 24: Rather than an `if` instead an `else` what about an `elsif`?

PR #545, opened 2020-08-28
Shards overrides

jwoertink:
This allows us to override shards during testing. By adding the override deps to the `shard.override.yml` file, any shard defined in generated projects including their dependencies will use this version instead. To do this, we have to use shards v0.12 which doesn't exist with the current latest crystal version (0.35.1). This means we have to build it from source. Once 1.0 is out, we will be able to remove that part.


.github/workflows/ci.yml
```diff
@@ -70,7 +80,9 @@ jobs:
             ${{ runner.os }}-build-
             ${{ runner.os }}-
       - name: Run tests
-        run: crystal spec
+        run: |
+          export SHARDS_OVERRIDE="$(pwd)/shard.override.yml"
```
jwoertink, line 84: I had to put this here because I couldn't get the dynamic `$(pwd)` to work under the `env` key, and it fails if you call `source` because that utility doesn't exist on this container for some reason (maybe security?).

paulcsmith: Do you know why this is needed at all? I thought it would use `shard.override.yml` by default? Sounds like it might be a bug in shards

script/override_shards
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+# source this file to export the variable
+
+export SHARDS_OVERRIDE="$(pwd)/shard.override.yml"
```
paulcsmith, line 5: Could you add a new line here?

shard.override.yml
```diff
@@ -0,0 +1,11 @@
+# This file is used to override the shards built by
+# generated Lucky apps. Comment these out if you want to
+# use the versions specified by these apps.
+
+# NOTE: https://github.com/crystal-lang/shards/issues/432
+# Until this is resolved, the default needs to be whatever the
+# default Lucky is.
+dependencies:
```
paulcsmith, line 8: I think it would be less error-prone to use this suggestion https://github.com/crystal-lang/shards/issues/432#issuecomment-683150150

I can see us running into confusion when we change shards.yml but forget to change the override.yml

.github/workflows/ci.yml
jwoertink: It does, but it assumes the file is in the same directory which in this case, it's not. We generate a test project where the `shard.yml` is in `test-project/shard.yml`, but the override is at `shard.override.yml` and not `test-project/shard.override.yml`. So in this case, we export the env var, and when we run `test-project/shards install`, it picks it up. Does that make sense?

paulcsmith: Ah yeah that makes sense! Thanks for the explanation!

PR #551, opened 2020-09-06
Revert DATABASE_URL rename to DB_URL

stephendolan:
This PR aims to make default-generated apps work with most DB services again
by renaming DB_URL back to DATABASE_URL.


paulcsmith (approved):
Thanks for catching this!

PR #555, opened 2020-09-25
Default to printing emails to logs in development

stephendolan:
This PR aims to make debugging emails a bit more straightforward for new Lucky users. I remember being quite confused the first time I tried to send an email in dev and ended up enabling SendGrid since I couldn't determine how to show the email content.

Since then, I've dug into the Carbon source and found the nifty `print_emails` option when initializing a `Carbon::DevAdapter`! I think it'd be handy to turn that on by default in dev, especially since if it's too noisy turning the flag off is pretty straightforward.

If this isn't agreeable, there are a few other options that could help (that I'm happy to open issues/PRs for):
- Add documentation for the `print_emails` flag to the Carbon readme
- Add documentation for the `print_emails` flag to the Lucky guides

### DIFF
--- a/src/web_app_skeleton/config/email.cr
+++ b/src/web_app_skeleton/config/email.cr
@@ -7,6 +7,8 @@ BaseEmail.configure do |settings|
     # If you do need emails, get a key from SendGrid and set an ENV variable
     send_grid_key = send_grid_key_from_env
     settings.adapter = Carbon::SendGridAdapter.new(api_key: send_grid_key)
+  elsif Lucky::Env.development?
+    settings.adapter = Carbon::DevAdapter.new(print_emails: true)
   else
     settings.adapter = Carbon::DevAdapter.new
   end


src/web_app_skeleton/config/email.cr
```diff
@@ -8,7 +8,7 @@ BaseEmail.configure do |settings|
     send_grid_key = send_grid_key_from_env
     settings.adapter = Carbon::SendGridAdapter.new(api_key: send_grid_key)
   else
-    settings.adapter = Carbon::DevAdapter.new
+    settings.adapter = Carbon::DevAdapter.new(print_emails: true)
```
jwoertink, line 11: I think it makes sense for this, but maybe turned off in test mode? Also, if we add a quick comment above this that's like `set this to false if logs become too noisy` or something? I'm horrible at wording stuff, but that would help. On a side note, I believe Paul may have mentioned that these will go in to breeze. If that's the case, then this could be a temporary deal anyway. In any case, I'm cool with it 👍

stephendolan: Ah, yes I was just adding this to Jumpstart and realized that the test output is **not** pretty, so I added an `elsif` for `development` and left the test environment silent!

jwoertink (approved):
Can you open an issue on the website to document this? I know those guides are lacking a lot, but it'll be nice to start tracking some of that.

paulcsmith (approved):
I think this is a good idea for now, but hopefully soonish we'll have Breeze and Pulsar hooked up so it will print a link to the logs that will tka you to a page with the email that was sent in dev!

stephendolan:
Will do, @jwoertink.

Can't wait, that sounds super convenient @paulcsmith!

PR #557, opened 2020-10-02
Task cli arg update

### LINKED ISSUE #556: CLI arg format error could use love
If you don't read Regex as a second language, then getting an error like:

> Invalid format for IndexPages. It should match /^[A-Z].+(?::+*)Page$/

can be a bit intimidating. 

One easy option could be to revamp the error in to an easy "paste" to play.crystal-lang so it looks like:

```
SomeTask was passed in invalid option format:

Be sure this condition returns true:

  "IndexPages".match?(/^[A-Z].+(?::+*)Page$/)

```

Or, another option would be to set an optional error message. Then the CLI arg becomes:

```crystal
positional_arg :page_class,
            "The name of the Page class. e.g. Users::IndexPage",
             format: /^[A-Z].+(?::+*)Page$/,
             error_message: "You did it wrong...."
```

jwoertink:
Fixes  #556 (or at least makes it a little more tolerable)

This PR does a few things for Tasks:

* Removed old deprecated `banner` macro
* Adds a new `output` property which is used for testing purposes
* Adds new `example` option to `arg` and `positional_arg` to help clarify the error output when a format is used

Currently with some tasks like [gen.page](https://github.com/luckyframework/lucky/blob/master/tasks/gen/page.cr#L20) we override the `call` method to take an option `IO` which is only used to test the output. The issue is that with the new CLI args, you have to pass them to `print_help_or_call` which calls `call`. With this `output` property, we can now set that to whatever we need (`IO::Memory.new`) and test that `our_task.output.to_s` contains whatever we're looking for.

### DIFF
--- a/spec/lucky_cli/task_spec.cr
+++ b/spec/lucky_cli/task_spec.cr
@@ -73,6 +73,12 @@ describe LuckyCli::Task do
       end
     end
 
+    it "provides an example for the error message on args" do
+      expect_raises(Exception, /Example: dark/) do
+        TaskWithRequiredFormatArgs.new.print_help_or_call(args: ["--theme=Spooky"])
+      end
+    end
+
     it "sets switch flags that default to false" do
       task = TaskWithSwitchFlags.new.print_help_or_call(args: [] of String).not_nil!
       task.admin?.should eq false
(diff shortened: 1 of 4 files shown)


spec/lucky_cli/task_spec.cr
```diff
@@ -27,16 +27,18 @@ describe LuckyCli::Task do
   describe "print_help_or_call" do
     it "prints the help_message for a found task when a help flag is passed" do
       %w(--help -h help).each do |help_arg|
-        io = IO::Memory.new
-        My::CoolTask.new.print_help_or_call(args: [help_arg], io: io)
-        io.to_s.should have_default_help_message
+        task = My::CoolTask.new
+        task.output = IO::Memory.new
```
paulcsmith, line 31: I like this idea 👍

paulcsmith (approved):
Love it!

jwoertink:
Sweet! Thanks for the Review!

PR #559, opened 2020-10-08
Generate default GitHub Actions config

### LINKED ISSUE #533: Use GitHub actions instead of Travis for generated apps
https://blog.roetzel.de/user/dave/post/using-github-actions-for-ci-of-lucky-apps

### LINKED ISSUE #558: Generate github action CI config
Generate github action config in same way as .travis.yml

If this is a good idea then I can make PR

skojin:
fixes #558 
fixes #533

### DIFF
--- a/src/web_app_skeleton/.github/workflows/ci.yml.ecr
+++ b/src/web_app_skeleton/.github/workflows/ci.yml.ecr
@@ -1,0 +1,114 @@
+name: <%= project_name %> CI
+
+on:
+  push:
+    branches: "*"
+  pull_request:
+    branches: "*"
+
+jobs:
+  check-format:
+    runs-on: ubuntu-latest
+    container:
+      image: crystallang/crystal:<%= Crystal::VERSION %>
+    steps:
+      - uses: actions/checkout@v2
+      - name: Format
+        run: crystal tool format --check
+
+  specs:
+    runs-on: ubuntu-latest
+    container:
+      image: crystallang/crystal:<%= Crystal::VERSION %>
+    env:
+      LUCKY_ENV: test
+      DB_HOST: postgres
+
+    services:
+      postgres:
+        image: postgres:12-alpine
+        env:
+          POSTGRES_PASSWORD: postgres
+        ports:
+          - 5432:5432
+       # Set health checks to wait until postgres has started
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install PostgreSQL client
+      run: |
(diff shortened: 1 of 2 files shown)


jwoertink (approved):
Yeah! This looks awesome!

src/web_app_skeleton/.github/workflows/ci.yml.ecr
```diff
@@ -0,0 +1,114 @@
...
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install PostgreSQL client
+      run: |
+        apt-get update
+        apt-get -yqq install libpq-dev postgresql-client
+<%- if browser? -%>
+    - name: Install Chromedriver
```
stephendolan, line 49: I don't think this step is required since the shard we use for webdrivers will install and add the appropriate binary to your home directory.

src/web_app_skeleton/.github/workflows/ci.yml.ecr
```diff
@@ -0,0 +1,114 @@
+name: <%= project_name %> CI
+
+on:
```
stephendolan, line 3: We can actually consolidate this to `on: [push, pull_request]`

src/web_app_skeleton/.github/workflows/ci.yml.ecr
```diff
@@ -0,0 +1,114 @@
...
+  push:
+    branches: "*"
+  pull_request:
+    branches: "*"
```
stephendolan, line 7: We should actually be able to consolidate this according to [the docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#on)

```suggestion
on: [push, pull_request]
```

src/web_app_skeleton/.github/workflows/ci.yml.ecr
stephendolan: Nevermind, it didn't pass without this :(

src/web_app_skeleton/.github/workflows/ci.yml.ecr
skojin: My version  is more flexible and self descriptive, and allow easy change "*" to master if need, e.g. like here
https://github.com/luckyframework/lucky_cli/blob/8be29ea4af81fb2ae027093ea8c38d56cb0cab11/.github/workflows/ci.yml#L4-L7

paulcsmith: That's a good point. I think keeping it as-is makes it easier to customize so can be left as-is

stephendolan: Sounds good to me!

PR #565, opened 2020-10-23
Install browser in CI instead of Chromedriver

### LINKED ISSUE #564: Update skeleton Github Actions ci.yml to only install chromium-browser
https://github.com/luckyframework/lucky_cli/blob/0efe5e465fb22555fa3b92c5dcb8d0eebd30fe62/src/web_app_skeleton/.github/workflows/ci.yml.ecr#L50

Right now the ci.yml file installs `chromium-chromedriver` but since generated apps have the `webdrivers.cr` shard, it can manage chromedrivers itself. A browser is needed though, so we should switch it to install `chromium-browser` and let the shard install additional dependencies where needed.

stephendolan:
Let webdrivers.cr install chromedriver for appropriate browser.

Closes #564 

Tested and verified here: https://github.com/stephendolan/lucky_jumpstart/runs/1298680502

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,8 +40,8 @@ jobs:
           apt-get update
           apt-get -yqq install libpq-dev postgresql-client
       - uses: actions/checkout@v2
-      - name: Install Chromedriver
-        run: apt-get -yqq install chromium-chromedriver
+      - name: Install browser
+        run: apt-get -yqq install chromium-browser
       - uses: actions/setup-node@v1
         with:
           node-version: "12.x"
--- a/src/web_app_skeleton/.github/workflows/ci.yml.ecr
+++ b/src/web_app_skeleton/.github/workflows/ci.yml.ecr
@@ -46,8 +46,8 @@ jobs:
         apt-get update
         apt-get -yqq install libpq-dev postgresql-client
 <%- if browser? -%>
-    - name: Install Chromedriver
-      run: apt-get -yqq install chromium-chromedriver
+    - name: Install browser
+      run: apt-get -yqq install chromium-browser
 
     - uses: actions/setup-node@v1
       with:


matthewmcgarvey (approved):
Thanks for making the PR 👍

.github/workflows/ci.yml
```diff
@@ -40,8 +40,8 @@ jobs:
           apt-get update
           apt-get -yqq install libpq-dev postgresql-client
       - uses: actions/checkout@v2
-      - name: Install Chromedriver
-        run: apt-get -yqq install chromium-chromedriver
+      - name: Install browser
+        run: apt-get -yqq install chromium-browser
```
jwoertink, line 44: But doesn't this mean that it's an extra install step? Because you need the driver either way, right?

stephendolan: Yeah, this would just dictate the **browser** we're using in the Docker container, letting the chromedriver shard pick the appropriate binary to download and test with from the chromedriver side.

At least, that's my understanding!

If we didn't run this in a Crystal docker container, the default GitHub actions container already has chromium installed. That said, then we'd have to install Crystal so it might be a wash.

matthewmcgarvey: Yes, so the code will later install the driver, but I believe this is the best decision because it will more closely match how users are/should be testing their applications.

stephendolan:
Ugh, this seems like a similar error to what I was getting here: https://github.com/luckyframework/lucky_cli/pull/553/checks?check_run_id=1144104086

Has anyone else had CI issues lately on this repo?

.github/workflows/ci.yml
stephendolan: I also realized I made this change in the actual CLI workflow file. Should still be relevant, but just FYI.

jwoertink (approved):
Once specs pass, you should be good. Can you also make an issue on the website to update our docs https://luckyframework.org/guides/testing/html-and-interactivity you'll see it mentions to install the driver directly. If we're going with "let the shard handle it", then we should have the docs just say something like "Install Chrome, or configure it to use X..."

stephendolan:
Oh no :( we have a flapping test I think. Something to look into later, but just keep it in mind if any of y'all have PRs that fail. I just clicked the manual "Re-run jobs" button and it passed the second time.

PR #566, opened 2020-10-23
Use Dependabot to automatically bump generated dependencies

### LINKED ISSUE #562: Automate generated app dependency updates using Dependabot
See https://github.com/luckyframework/lucky/discussions/1261

Basically, we want to see if we can set up Dependabot to send us PRs for the generated app dependencies in our scaffolded templates.

stephendolan:
Closes #562 

Note that I've set myself as the reviewer for these to hopefully reduce the anxiety any other team members have about the maintenance burden.

I'll revisit in a few months to see if we should just add the entire team as the default "reviewer" field.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,19 @@
+version: 2
+
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    target-branch: master
+    reviewers:
+      - stephendolan
+
+  - package-ecosystem: npm
+    directory: "/src/browser-app-skeleton"
+    schedule:
+      interval: daily
+    target-branch: master
+    versioning-strategy: increase
+    reviewers:
+      - stephendolan


stephendolan:
Converting this to a draft until I get https://github.com/luckyframework/lucky_cli/pull/553 figured out.

stephendolan:
#553 is in! This may be one of those where we have to do a couple tweaks after it's in master based on how PRs are created, but I _think_ this will do what we want.

.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
...
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    target-branch: master
+    reviewers:
+      - stephendolan
+
+  - package-ecosystem: npm
+    directory: "/src/browser-app-skeleton"
```
jwoertink, line 13: I'm surprised at how this path is written. As opposed to being `src/browser-app-skeleton`. Does this automatically prepend `.` or something?

.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
...
+    schedule:
+      interval: daily
```
jwoertink, line 15: Would it be less hectic if we did like monthly instead?

jwoertink (commented):
I'm not too sure what all is going on here. I've never used dependabot directly. I'm assuming we can try this out a bit and see how it goes. Worst case is we just rip it out later if it's not working for us 🤔

.github/dependabot.yml
stephendolan: Perhaps... the trouble is that doing it monthly doesn't mean you get one combined PR with the dependency updates, it just means that you get a month's worth of PRs all at once.

If we leave it daily, there's no harm in ignoring them until we've got time, and Dependabot will automatically close and open new ones if dependency versions are changed after a PR is created.

.github/dependabot.yml
stephendolan: Now you've got me wondering... their docs made me think the slash was necessary, but I've never not had something in the root directory to test it.

https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuration-options-for-dependency-updates#directory

.github/dependabot.yml
matthewmcgarvey: After going down the weekly route, it's way less hectic to receive a PR every couple of days than receiving 15 (I'm pretty sure that's the default limit) PR's all out once.

stephendolan:
Going to merge this one in and see what happens! We can always back the file out, but there are some things I'm just not sure about until it's live.

PR #571, opened 2020-11-01
Fix init.custom task name in banner text

ZimbiX:
Running `lucky custom.init` does not work.

https://github.com/luckyframework/lucky_cli/blob/13126800b2b137c9d704950311425393a7b3ea92/src/lucky.cr#L37-L38


jwoertink (changes requested):
oops! 😅  pretty sure that's my bad. We should rename the file and class too so that way it all matches. Can you rename this to `init_custom.cr` and `LuckyCli::InitCustom`? It'll also need updating the require [here](https://github.com/luckyframework/lucky_cli/blob/master/src/lucky_cli.cr#L5) Thanks for catching this!

ZimbiX:
Fair enough - done =)

jwoertink (approved):
Specs on this repo are wonky right now so I'm not too worried about that, but thanks for doing this!

PR #552, opened 2020-09-17
Rename with_defaults to tag_defaults

stephendolan:
This PR aims to close #548

It will need to be released in conjunction with https://github.com/luckyframework/lucky/pull/1262

This PR tackles:
- Renaming with_defaults in `browser_app_skeleton/src/components/shared/field.cr` to the new `tag_defaults` signature implemented [here](https://github.com/luckyframework/lucky/pull/1262)
- Renaming the yielded block from `input_builder` to `tag_builder` to keep all Lucky-generated docs/code/comments consistent and (hopefully) better convey meaning

I wasn't sure how to get the specs passing here since the method won't be found in Lucky until a new version is available.

PR #572, opened 2020-11-01
lucky dev: When Ctrl+C is pressed, wait for the child process to handle the signal and finish - rather than us immediately exiting and leaving it to log shutting down messages from the background

### LINKED ISSUE #570: lucky dev Ctrl+C does not wait until the process manager exits
When I run `lucky dev` and at any point press Ctrl+C to stop it, sometimes the output is unclean, still logging process exits after my next terminal prompt has shown. This seems to be due to `lucky dev` not waiting for the process manager to exit when it receives the `SIGINT` signal.

It's most evident when using Foreman, since it's a bit slower to react to Ctrl+C than Overmind. Without a way to select the process manager in use, I've been forcing use of `foreman` by disabling my binaries which have higher priority:

```bash
sudo mv /usr/bin/overmind /usr/bin/overmindx
sudo mv /usr/local/bin/heroku /usr/local/bin/herokux
```

Example:

Screenshot from 2020-11-01 17-25-06

It looks like a fix would involve catching the interrupt before running the process manager

```
Signal::INT.trap do
  puts "Int"
end

Process.run("foreman", ["start", "-f", "Procfile.dev"], output: STDOUT, error: STDERR)
```

which works:

Screenshot from 2020-11-01 17-28-45

I reckon this could be added here:

https://github.com/luckyframework/lucky_cli/blob/13126800b2b137c9d704950311425393a7b3ea92/src/lucky_cli/process_runner.cr#L33-L38

I'll start working on a PR.

ZimbiX:
Resolves #570. I've just used the approach I suggested there.

I wasn't sure how/if to write a spec for this, so I haven't. Is that acceptable?

To test it manually, I ran this from my app directory before and after the change to verify the behaviour with Foreman:

```bash
$ crystal run ~/Projects/lucky_cli/src/lucky.cr -- dev
```

Before:

Screenshot from 2020-11-02 01-43-04

After:

Screenshot from 2020-11-02 01-43-30

----

The slow already-failing integration specs were an unfortunately painful distraction =P

The first:

```
In src/components/shared/field.cr:46:5

 46 | tag_defaults field: attribute do |tag_builder|
      ^-----------
Error: undefined local variable or method 'tag_defaults' for Shared::Field(String | Nil)
```

Finding #552, I learned I needed to be using lucky master. But was then confused why `shard.override.yml` didn't seem to be working to select it. I finally realised I had to use enable that by running the spec like:

```bash
(. ./script/override_shards && crystal spec spec/integration/init_web_spec.cr:10)
```

Running that then resulted in:

```
In lib/avram/src/avram/operation.cr:81:16

 81 | abstract def run
                   ^--
Error: abstract `def Avram::Operation#run()` must be implemented by RequestPasswordReset
```

From which I found https://github.com/luckyframework/avram/pull/469. I briefly considered updating `RequestPasswordReset` to this new interface, but couldn't easily see how to do so. At that point I backed away slowly, haha. I'll leave you folks to work out the inter-repo dependency in rolling out that change.


jwoertink (commented):
Thanks for the explanations and screenshots. I think this change is fine. We are in the midst of a bit of remodeling, so sorry for the dust 😅  we will work out the weird dependency issues this week and get that all set before merging this in.

src/lucky_cli/process_runner.cr
```diff
@@ -31,6 +31,9 @@ module LuckyCli
     end
 
     private def run(process, args)
+      # When Ctrl+C is pressed, wait for the child process to handle the signal and finish - rather than us immediately exiting and leaving it to log shutting down messages from the background
```
jwoertink, line 34:
```suggestion
      # When Ctrl+C is pressed, wait for the child process to handle the signal and finish - 
      # rather than us immediately exiting and leaving it to log "shutting down" messages from the background
```

jwoertink (commented):
Ok, hopefully these specs pass now. 🤞

jwoertink (approved):
Sweet! I tested this locally, and it works great.

PR #573, opened 2020-11-01
Disable 'Compiling' spinner on CI to avoid clogging logs

### LINKED ISSUE #560: Hide "Compiling" spinner if in CI
If the `CI` ENV var is set, we should not run the spinner. Right now it's outputting a line for every frame and clogging log output

ZimbiX:
Resolves #560.

The Spinner class is complex, so I didn't really want to add to the complexity - especially with no tests. I considered implementing a `NullSpinner`, but the spinner interface is not super simple; it felt simpler to use the existing spinner class conditionally from its one usage in `BuildAndRunTask`.

It is curious that the spinner supports several unused features.

Demonstration:

Screenshot from 2020-11-02 02-27-06


jwoertink (commented):
Nice! This will for sure make it easier to read the CI output 😂  I'll get working on the updates to fix the spec issues, but this look good to me 👍

jwoertink (approved):
So much easier to scroll though those log outputs! 😂  thanks for this!

PR #575, opened 2020-11-02
Using Avram master to update all generated Operations to use the new structure

### LINKED ISSUE #574: Update generated Operations to new structure
With https://github.com/luckyframework/avram/pull/469 being merged in, Operations now have a new interface. We need to make sure all of the generated operations are updated to use these.

jwoertink:
Fixes #574

### DIFF
--- a/shard.override.yml
+++ b/shard.override.yml
@@ -9,3 +9,6 @@ dependencies:
   lucky:
     github: luckyframework/lucky
     branch: master
+  avram:
+    github: luckyframework/avram
+    branch: master
--- a/shard.yml
+++ b/shard.yml
@@ -16,4 +16,5 @@ dependencies:
 development_dependencies:
   ameba:
     github: crystal-ameba/ameba
-    version: ~> 0.13
+    branch: master
+    #version: ~> 0.13
--- a/spec/integration/init_web_spec.cr
+++ b/spec/integration/init_web_spec.cr
@@ -137,16 +137,16 @@ end
 private def with_project_cleanup(project_directory = "test-project", skip_db_drop = false)
   yield
 
-  FileUtils.cd project_directory do
+  FileUtils.cd(project_directory) {
     output = IO::Memory.new
     Process.run(
       "lucky db.drop",
       output: output,
       shell: true
-    ) unless skip_db_drop
+    )
 
     output.to_s.should contain("Done dropping")
-  end
+   } unless skip_db_drop
 ensure
   FileUtils.rm_rf project_directory
 end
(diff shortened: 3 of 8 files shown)


jwoertink:
this spec error doesn't even make sense... is the CI broke? 🤔

wontruefree:
should we cut a point release for avram and use that here?

jwoertink:
Not yet just because there's still a few more items to round out this next release. Pointing to master should be fine for now, but it's weird that the CI is failing on that one spec. My guess is it's unrelated. 🤷

wontruefree:
I dont know this project as well but if someone updates luck_cli and then never re downloads it what happens?  It will just pull master as of now?

jwoertink:
That's mostly going to come from https://github.com/luckyframework/lucky/blob/master/shard.yml#L34 as far as the task components goes. Otherwise it's just whatever version you have installed. If you install from master, then you'd have to update on your own as new updates come in. But for now that shouldn't really affect anyone until we do a point release on here.

wontruefree:
Seems good.  Thanks for the thorough response :)

matthewmcgarvey:
So the error from the failing spec is:
```
Unhandled exception: Unexpected error while running migrations: There was a problem running this statement:

  CREATE TABLE users (
  id bigserial PRIMARY KEY,
  created_at timestamptz NOT NULL,
  updated_at timestamptz NOT NULL,
  email text NOT NULL,
  encrypted_password text NOT NULL);
CREATE UNIQUE INDEX users_email_index ON users USING btree (email);

Problem:

  relation "users" already exists (Exception)
  from lib/avram/src/avram/migrator/migrator.cr:7:5 in 'call'
  from lib/avram/src/avram/tasks/db/migrate.cr:3:1 in 'print_help_or_call'
  from lib/lucky_cli/src/lucky_cli/runner.cr:27:9 in 'run'
  from tasks.cr:20:1 in '__crystal_main'
  from ../../../../usr/share/crystal/src/crystal/main.cr:105:5 in 'main_user_code'
  from ../../../../usr/share/crystal/src/crystal/main.cr:91:7 in 'main'
  from ../../../../usr/share/crystal/src/crystal/main.cr:114:3 in 'main'
  from __libc_start_main
  from _start
  from ???
```

@jwoertink Is that what is weird? Looking at past specs, it seems like the integration specs are a bit flakey

jwoertink:
Yeah. I was just confused why it's all of a sudden failing. Like, is it a race condition? Or does this PR actually break something?  I'll take a look at it later when I have a bit more time. Unless you notice something @matthewmcgarvey

matthewmcgarvey (approved):

jwoertink:
Sweet baby cheeseus! it's passing 😂

PR #576, opened 2020-11-04
Add `task` env detection

KCErb:
I think it'd be nice to know that a task runner is calling your app code sometimes. So these changes add an env var to the entrypoint `tasks.cr` and a helper in the env config.

My specific use case is that I have some code I want to run during server startup, but not during `db.migrate` and friends. This PR will allow me to guard that code like so:

```crystal
Something.init unless Lucky::Env.task?
```

### DIFF
--- a/src/web_app_skeleton/config/env.cr
+++ b/src/web_app_skeleton/config/env.cr
@@ -10,4 +10,8 @@ module Lucky::Env
   def name
     ENV["LUCKY_ENV"]? || "development"
   end
+  
+  def task?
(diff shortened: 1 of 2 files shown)


jwoertink (commented):
I like it 👍  I think you'll need to run ameba or the formatter locally so that part passes, but the specs are gonna fail until we get a few other things in.

jwoertink:
Ok, you'll have to rebase, but then specs will be good 👍

jwoertink (approved):

PR #579, opened 2020-11-05
New cleanup method for specs

### LINKED ISSUE #578: Error when running specs
```
Unhandled exception: Unexpected error while running migrations: There was a problem running this statement:

  CREATE TABLE users (
  id bigserial PRIMARY KEY,
  created_at timestamptz NOT NULL,
  updated_at timestamptz NOT NULL,
  email text NOT NULL,
  encrypted_password text NOT NULL);
CREATE UNIQUE INDEX users_email_index ON users USING btree (email);

Problem:

  relation "users" already exists (Exception)
  from lib/avram/src/avram/migrator/migrator.cr:7:5 in 'call'
  from spec/setup/setup_database.cr:2:1 in '__crystal_main'
  from ../../../../../../usr/local/Cellar/crystal/0.35.1_1/src/crystal/main.cr:105:5 in 'main_user_code'
  from ../../../../../../usr/local/Cellar/crystal/0.35.1_1/src/crystal/main.cr:91:7 in 'main'
  from ../../../../../../usr/local/Cellar/crystal/0.35.1_1/src/crystal/main.cr:114:3 in 'main'
```

This may be a recent change related to Avram master branch, but I just noticed that I now get this locally as well. I think we just need to make sure that we're clearing out the `test-project` fully between integration specs to include resetting the DB that gets generated.

jwoertink:
Fixes #578

🐔  and 🥚  time.... Specs won't pass here since it needs #574 but that one won't pass until we get this one in 😬

### DIFF
--- a/spec/integration/init_web_spec.cr
+++ b/spec/integration/init_web_spec.cr
@@ -2,72 +2,85 @@ require "../spec_helper"
 
 include ShouldRunSuccessfully
 
-Spec.before_each do
-  FileUtils.rm_rf("test-project")
-end
-
 describe "Initializing a new web project" do
   it "creates a full web app successfully" do
     puts "Web app: Running integration spec. This might take awhile...".colorize(:yellow)
-    should_run_successfully "crystal run src/lucky.cr -- init.custom test-project"
-    FileUtils.cp("spec/support/cat.gif", "test-project/public/assets/images/")
+    with_project_cleanup do
+      should_run_successfully "crystal run src/lucky.cr -- init.custom test-project"
+      FileUtils.cp("spec/support/cat.gif", "test-project/public/assets/images/")
 
-    File.delete("test-project/.env")
-    compile_and_run_specs_on_test_project
-    File.read("test-project/.travis.yml").should contain "postgresql"
-    File.read(".github/workflows/ci.yml").should contain "postgres"
-    File.read("test-project/public/mix-manifest.json").should contain "images/cat.gif"
-    File.exists?("test-project/public/favicon.ico").should eq true
-    File.exists?("test-project/.env").should eq true
+      # Testing that dashes are translated to underscores
+      shard_yml_name = File.read_lines("test-project/shard.yml").find { |line| line =~ /^name:/ }
+      shard_yml_name.should eq("name: test_project")
+      File.exists?("test-project/src/test_project.cr").should be_true
+
+      File.delete("test-project/.env")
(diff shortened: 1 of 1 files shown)


spec/integration/init_web_spec.cr
```diff
@@ -106,20 +119,6 @@ describe "Initializing a new web project" do
...
-    Process.run(
-      "crystal run src/lucky.cr -- init.custom test-project",
-      env: ENV.to_h,
-      output: output,
-      shell: true
-    )
-
-    shard_yml_name = File.read_lines("test-project/shard.yml").find { |line| line =~ /^name:/ }
-    shard_yml_name.should eq("name: test_project")
-    File.exists?("test-project/src/test_project.cr").should be_true
```
jwoertink, line 121: This is still tested, it's just moved to the first spec block. These specs are pretty expensive to run, and since all of the integration specs use the same `test-project`, we can piggyback off the first spec to ensure the name is translated from dashes to underscores, and save some time by generating 1 less project.

spec/integration/init_web_spec.cr
```diff
@@ -133,3 +132,18 @@ private def compile_and_run_specs_on_test_project
...
+  output = IO::Memory.new
+  Process.run(
+    "lucky db.drop",
+    env: ENV.to_h,
+    output: output,
+    shell: true
+  )
+  output.to_s.should contain("Done dropping")
+  FileUtils.cd root_directory
+  FileUtils.rm_rf project_directory
```
jwoertink, line 148: This was moved from the `Spec.before_each` block. It should only need to run after any spec that creates the project directory.

spec/integration/init_web_spec.cr
```diff
@@ -2,72 +2,85 @@ require "../spec_helper"
 
 include ShouldRunSuccessfully
 
-Spec.before_each do
```
jwoertink, line 5: I believe that `Spec.before_each` isn't scoped, so even though it was added in this file, I'm assuming it's being run on ALL specs for the entire project. (unless that's been fixed in Crystal?)

matthewmcgarvey (approved):
This seems fine as a change. Just a note, though. It will only cleanup, right now, if the test passes. An error will mean that none of it runs.

jwoertink:
hmmm.... Good point... maybe I should do something like...

```crystal
begin
  yield
  # ..
ensure
  FileUtils.rm_rf project_directory
end
```

?

PR #586, opened 2020-12-09
Remove unnecessary .gitignore additions - start_server/*.dwarf

robacarp:
To stay in step with https://github.com/luckyframework/lucky/pull/1358

start_server and start_server.dwarf are now emitted to bin/
which is already ignored by the crystal ignore template.

### DIFF
--- a/src/generators/web.cr
+++ b/src/generators/web.cr
@@ -66,8 +66,6 @@ class LuckyCli::Generators::Web
 
   private def setup_gitignore
     append_text to: ".gitignore", text: <<-TEXT
-    start_server
-    *.dwarf
     *.local.cr
     .env
     /tmp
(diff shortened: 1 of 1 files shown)


src/generators/web.cr
```diff
@@ -152,8 +150,7 @@ class LuckyCli::Generators::Web
     append_text to: "shard.yml", text: <<-DEPS_LIST
     dependencies:
       lucky:
-        github: luckyframework/lucky
-        version: ~> 0.24.0
+        path: /Users/robert/Documents/repositories/luckyframework/lucky
```
matthewmcgarvey, line 153: 🙈 This should be undone.

matthewmcgarvey (changes requested):

src/generators/web.cr
```diff
@@ -66,8 +66,6 @@ class LuckyCli::Generators::Web
 
   private def setup_gitignore
     append_text to: ".gitignore", text: <<-TEXT
-    start_server
-    *.dwarf
```
jwoertink, line 70: I think we can actually just keep this. In my apps, I can't run `lucky dev` (technical reasons), so I actually just run the build myself and if I forget to send it to the bin, it'll leave a .dwarf file.

src/generators/web.cr
robacarp: okay but how will you know where I store my code 😆

robacarp:
If we're going to leave the .gitignore, there's nothing to keep in here! Thanks for looking.

PR #592, opened 2020-12-18
Upgrade to Laravel Mix 6

stephendolan:
Laravel Mix 6 dropped today!

I've been running the beta in 3 production apps for the last few weeks, so I believe it's relatively safe to include in our next CLI release for all generated apps.

It also significantly reduces the code in our `package.json` scripts, which might be a bit mysterious behind the `mix` abstraction, but I see as a definite bonus personally.

Here are the release notes:
https://github.com/JeffreyWay/laravel-mix/blob/master/CHANGELOG.md#60

### DIFF
--- a/src/browser_app_skeleton/package.json
+++ b/src/browser_app_skeleton/package.json
@@ -8,18 +8,18 @@
   },
   "scripts": {
     "heroku-postbuild": "yarn prod",
-    "dev": "yarn run webpack --progress --hide-modules --color --config=node_modules/laravel-mix/setup/webpack.config.js",
-    "watch": "yarn run webpack --watch --hide-modules --color --config=node_modules/laravel-mix/setup/webpack.config.js",
-    "prod": "NODE_ENV=production yarn run webpack --progress --hide-modules --color --config=node_modules/laravel-mix/setup/webpack.config.js"
+    "dev": "mix",
+    "watch": "mix watch",
+    "prod": "mix --production"
   },
   "devDependencies": {
     "@babel/compat-data": "^7.9.0",
     "browser-sync": "^2.26.7",
-    "compression-webpack-plugin": "^6.0.1",
-    "laravel-mix": "^5.0.5",
+    "compression-webpack-plugin": "^7.0.0",
+    "laravel-mix": "^6.0.0",
+    "postcss": "^8.1.0",
     "resolve-url-loader": "^3.1.1",
     "sass": "^1.26.10",
-    "sass-loader": "^10.0.2",
-    "vue-template-compiler": "^2.6.12"
+    "sass-loader": "^10.0.2"
   }
 }
--- a/src/browser_app_skeleton/webpack.mix.js
+++ b/src/browser_app_skeleton/webpack.mix.js
@@ -46,7 +46,7 @@ mix
   // https://github.com/JeffreyWay/laravel-mix/blob/master/docs/mixjs.md
   .js("src/js/app.js", "public/js")
   // SASS entry file. Uses autoprefixer automatically.
-  .sass("src/css/app.scss", "public/css")
+  .sass("src/css/app.scss", "css")
   // Customize postCSS:
   // https://github.com/JeffreyWay/laravel-mix/blob/master/docs/css-preprocessors.md#postcss-plugins
   .options({
@@ -73,7 +73,6 @@ mix
 
 // Full API
 // Docs: https://github.com/JeffreyWay/laravel-mix/tree/master/docs#readme
-// Source: node_modules/laravel-mix/setup/webpack.mix.js
 //
 // mix.js(src, output);
 // mix.react(src, output); <-- Identical to mix.js(), but registers React Babel compilation.
--- a/src/browser_app_skeleton/webpack.mix.js
+++ b/src/browser_app_skeleton/webpack.mix.js
@@ -27,7 +27,7 @@ if (mix.inProduction()) {
   })
   plugins.push(gzipCompression)
 
-  // Add additional compression plugins here. 
+  // Add additional compression plugins here.
   // For example if you want to add Brotli compression:
   //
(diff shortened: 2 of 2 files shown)


src/browser_app_skeleton/package.json
```diff
@@ -8,18 +8,18 @@
...
     "browser-sync": "^2.26.7",
-    "compression-webpack-plugin": "^6.0.1",
-    "laravel-mix": "^5.0.5",
+    "compression-webpack-plugin": "^7.0.0",
+    "laravel-mix": "^6.0.0",
+    "postcss": "^8.1.0",
     "resolve-url-loader": "^3.1.1",
     "sass": "^1.26.10",
-    "sass-loader": "^10.0.2",
-    "vue-template-compiler": "^2.6.12"
```
stephendolan, line 23: We can remove this now that the Laravel Mix team has removed the assumption that every app will use Vue 🎉

src/browser_app_skeleton/package.json
```diff
@@ -8,18 +8,18 @@
...
+    "prod": "mix --production"
   },
   "devDependencies": {
     "@babel/compat-data": "^7.9.0",
     "browser-sync": "^2.26.7",
-    "compression-webpack-plugin": "^6.0.1",
-    "laravel-mix": "^5.0.5",
+    "compression-webpack-plugin": "^7.0.0",
+    "laravel-mix": "^6.0.0",
+    "postcss": "^8.1.0",
```
stephendolan, line 20: PostCSS v8 is a new Laravel Mix 6 prerequisite

src/browser_app_skeleton/package.json
```diff
@@ -8,18 +8,18 @@
...
+    "dev": "mix",
+    "watch": "mix watch",
+    "prod": "mix --production"
   },
   "devDependencies": {
     "@babel/compat-data": "^7.9.0",
     "browser-sync": "^2.26.7",
-    "compression-webpack-plugin": "^6.0.1",
-    "laravel-mix": "^5.0.5",
+    "compression-webpack-plugin": "^7.0.0",
```
stephendolan, line 18: Webpack 4 and Compression Webpack Plugin 7 weren't playing nice together, so this can be bumped now as well.

src/browser_app_skeleton/webpack.mix.js
```diff
@@ -46,7 +46,7 @@ mix
   // https://github.com/JeffreyWay/laravel-mix/blob/master/docs/mixjs.md
   .js("src/js/app.js", "public/js")
   // SASS entry file. Uses autoprefixer automatically.
-  .sass("src/css/app.scss", "public/css")
+  .sass("src/css/app.scss", "css")
```
stephendolan, line 49: This still outputs to the `public` folder, but seems to assume that namespace now.

Leaving this in its older form yields files in `public/public/css`

src/browser_app_skeleton/webpack.mix.js
```diff
@@ -73,7 +73,6 @@ mix
 
 // Full API
 // Docs: https://github.com/JeffreyWay/laravel-mix/tree/master/docs#readme
-// Source: node_modules/laravel-mix/setup/webpack.mix.js
```
stephendolan, line 76: This is no longer generated for Laravel Mix 6 😢

src/browser_app_skeleton/package.json
```diff
@@ -8,18 +8,18 @@
   },
   "scripts": {
     "heroku-postbuild": "yarn prod",
-    "dev": "yarn run webpack --progress --hide-modules --color --config=node_modules/laravel-mix/setup/webpack.config.js",
-    "watch": "yarn run webpack --watch --hide-modules --color --config=node_modules/laravel-mix/setup/webpack.config.js",
-    "prod": "NODE_ENV=production yarn run webpack --progress --hide-modules --color --config=node_modules/laravel-mix/setup/webpack.config.js"
+    "dev": "mix",
```
jwoertink, line 11: When you run `yarn` on a new lucky project, will you get a global `mix` executable? Or does this need to be `yarn run mix` so it scopes to this project?

src/browser_app_skeleton/package.json
jwoertink: That's awesome.

src/browser_app_skeleton/webpack.mix.js
jwoertink: What about the `js` one? That still says `public/js` 🤔 . Also, is there a config for this? If so, maybe we can expose the config by default and set it to public. Then people won't have to go digging through docs to change that (if they needed).

jwoertink: oh wait, line 60 says `.setPublicPath("public")`, so maybe that's where it's configured. If that's so, then I think just the js part needs to be updated, right?

jwoertink (commented):
Glad you're keeping up with this! I don't know anything about any of this 😂

src/browser_app_skeleton/webpack.mix.js
stephendolan: Weirdly enough, the JS part outputs correctly! I think moving that `setPublicPath` up will work, though, so I'll run some tests and re-submit so that it's a bit easier to reason about.

src/browser_app_skeleton/package.json
stephendolan: `yarn install` was sufficient for me. I think `package.json` scripts have `node_modules/bin` in their path.

src/browser_app_skeleton/webpack.mix.js
stephendolan: Okay, tested this locally and it worked when I moved up the `setPublicPath` call! Good catch!

jwoertink (approved):
This looks good to me! I hope it magically fixes my random serviceworker issue 😂

Issue #593, opened 2020-12-26
Remove usages of the route macro


matthewmcgarvey:
Replace it with explicit request method and path

Connected to https://github.com/luckyframework/lucky/discussions/789

PR #594, opened 2020-12-26
Removes all uses of route macro from newly generated apps

matthewmcgarvey:
Fixes https://github.com/luckyframework/lucky_cli/issues/593


jwoertink (approved):
Sweet!

PR #601, opened 2021-02-01
Rename procfile target

robacarp:
This is a compliment to https://github.com/luckyframework/heroku-buildpack-crystal/pull/11

See https://github.com/luckyframework/heroku-buildpack-crystal/pull/11#issuecomment-770074711

I also added a `target` stanza to the shard.yml. It's an out of context update and I'm happy to pull the commit if needed.

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -4,6 +4,10 @@ version: 0.25.0
 authors:
   - Paul Smith <paulcsmith0218@gmail.com>
 
+targets:
+  lucky:
+    main: src/lucky.cr
+
 crystal: 0.35.1
 
 license: MIT
--- a/src/web_app_skeleton/Procfile.ecr
+++ b/src/web_app_skeleton/Procfile.ecr
@@ -1,0 +1,2 @@
+web: bin/<%= project_name %>
+release: lucky db.migrate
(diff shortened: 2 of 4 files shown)


src/web_app_skeleton/Procfile.ecr
```diff
@@ -0,0 +1,2 @@
+web: bin/<%= project_name %>
```
jwoertink, line 1: Let's add a spec line https://github.com/luckyframework/lucky_cli/blob/master/spec/integration/init_web_spec.cr#L19 that looks for `Procfile` to contain `bin/test-project`

shard.yml
```diff
@@ -4,6 +4,10 @@ version: 0.25.0
 authors:
   - Paul Smith <paulcsmith0218@gmail.com>
 
+targets:
+  lucky:
+    main: src/lucky.cr
```
jwoertink, line 9: 👍

jwoertink (changes requested):
I'm not sure if I've ever seen the deploy to heroku specs run or not.. Will be interesting to see if they run and pass. I think just a little spec to ensure that's good, and we can merge this in.

src/web_app_skeleton/Procfile.ecr
robacarp: Good idea!

jwoertink (approved):
Let's roll with it!

PR #605, opened 2021-02-08
Bump generated Carbon dependency version to 0.1.3

stephendolan:
This may have been intentional, but it appears as though we released an 0.36.1 compatible release for Carbon that we're not leveraging by default for generated apps.

Here's the Carbon commit:
https://github.com/luckyframework/carbon/commit/e36bcb35a77b64cfe297dd96e9854d8d510657ae

### DIFF
--- a/src/generators/web.cr
+++ b/src/generators/web.cr
@@ -159,7 +159,7 @@ class LuckyCli::Generators::Web
         version: ~> 0.7.2
       carbon:
         github: luckyframework/carbon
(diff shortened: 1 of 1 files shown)


jwoertink:
Ah, this wasn't totally necessary just because the `~>` would install the patch update anyway. You can merge this in if you'd like, but I don't think a new release would be necessary.

stephendolan:
Oh, right! Yeah, I'm just going to close this out.

PR #619, opened 2021-04-02
Run github actions outside of docker container

matthewmcgarvey:
If this works, it will simplify the yml and speed up the jobs a bit


jwoertink:
> If this works

*narrator's voice* it didn't work

jwoertink (approved):
Sweet!

Issue #620, opened 2021-04-03
Using shard for your own tasks includes a lot of unwanted code


matthewmcgarvey:
If I have a library and want to add custom tasks, I have to have a dependency on lucky_cli. Makes sense but it includes everything that is used for the CLI as well as the classes to build tasks. One of those things is the `teeplate` dependency that some of the tasks use to generate new projects ([example](https://github.com/luckyframework/lucky_cli/blob/be0369c1a30a058d35b06b091401f22aebaf8222/src/lucky_cli/api_authentication_template.cr#L2)). There's no reason to ship that dependency to user's of the library and it has downsides. On an uncached compile, a [specific macro](https://github.com/luckyframework/teeplate/blob/797fc0c5414d8eff06b6d76c04d9be881623a636/src/lib/file_tree.cr#L7-L9) in teeplate takes 10 seconds to run. So now every shard that uses lucky_cli inherits that 10 second penalty.

```bash
❯ crystal build -s --release src/lucky_cli.cr
Parse:                             00:00:00.000088009 (   0.75MB)
Semantic (top level):              00:00:09.789230517 ( 137.86MB)
Semantic (new):                    00:00:00.001426725 ( 137.86MB)
Semantic (type declarations):      00:00:00.013700447 ( 137.86MB)
Semantic (abstract def check):     00:00:00.003392430 ( 137.86MB)
Semantic (ivars initializers):     00:00:00.011461789 ( 153.86MB)
Semantic (cvars initializers):     00:00:00.231323215 ( 153.98MB)
Semantic (main):                   00:00:00.030838551 ( 153.98MB)
Semantic (cleanup):                00:00:00.003465031 ( 153.98MB)
Semantic (recursive struct check): 00:00:00.000615239 ( 153.98MB)
Codegen (crystal):                 00:00:00.266971116 ( 169.98MB)
Codegen (bc+obj):                  00:00:08.413800449 ( 169.98MB)
Codegen (linking):                 00:00:00.161235795 ( 169.98MB)
dsymutil:                          00:00:00.040868893 ( 169.98MB)

Macro runs:
 - /Users/matthew/workspace/crystal/lucky_cli/lib/teeplate/src/lib/file_tree/macros/directory.cr: 00:00:09.360855688

Codegen (bc+obj):
 - no previous .o files were reused
```

[...]

PR #621, opened 2021-04-03
Make sure library code is completely separated from cli code

matthewmcgarvey:
Fixes https://github.com/luckyframework/lucky_cli/issues/620

This is a pretty big change, what's the point?

Well as #620 points out, the [teeplate shard](https://github.com/luckyframework/teeplate/blob/797fc0c5414d8eff06b6d76c04d9be881623a636/src/lib/file_tree.cr#L7-L9) has some code that causes compilation to slow down by 10 seconds when it can't use the cache. All that for code that doesn't actually need to run except for when the shard is built to be run as a cli.

Almost all of the code movement is because I want to make sure we never accidentally include the cli code in the library code going forward.


matthewmcgarvey:
Before approving/merging this, let's consider pulling the task code into a separate shard so that we avoid circular dependencies

src/cli/templates/browser_authentication_src_template (Copy 1).cr
```diff
@@ -0,0 +1,3 @@
+class BrowserAuthenticationSrcTemplate < Teeplate::FileTree
```
jwoertink, line 1: The filename on this says "Copy 1"

jwoertink:
wow. intense 😂  The end result will be worth it though. Since this is sort of a big thing, let me know what you'd like me to do. I don't want to just hop in and step on your toes. If you want me to look at breaking out the tasks deal, I can start there. Or unless you've already started on that..

matthewmcgarvey:
Closing in favor of https://github.com/luckyframework/lucky_cli/pull/622

PR #622, opened 2021-04-04
Replace task code with LuckyTask

matthewmcgarvey:
Fixes https://github.com/luckyframework/lucky_cli/issues/620

In https://github.com/luckyframework/lucky_cli/pull/621 I experimented with completely separating the CLI code from the task code but it was pointed out that we could extract a shard and that would possibly solve another problem of circular dependencies.

So I have made a [LuckyTask](https://github.com/luckyframework/lucky_task) and incorporated it into this library and removed the existing task code. This would be a bigger breaking change as we'd have to make sure we update all the places tasks are made in various libraries.


matthewmcgarvey:
In order for the specs to pass, Avram will have to be updated first

matthewmcgarvey:
There's one spec that uses lucky tasks so it will have to be updated as well before the last spec will pass

jwoertink (approved):
Is there anything else left for this move over?

matthewmcgarvey:
@jwoertink Nope!

Issue #640, opened 2021-06-09
normalize.scss dependency produces a warning


matthewmcgarvey:
https://github.com/JohnAlbin/normalize-scss is no longer maintained and a a new project generated produces a warning on startup

```bash
8:34:02 PM assets.1       |  DEPRECATION WARNING
8:34:02 PM assets.1       |  : Using / for division is deprecated and will be removed in Dart Sass 2.0.0.
8:34:02 PM assets.1       |  Recommendation: math.div($value, $relative-to)
8:34:02 PM assets.1       |  More info and automated migrator: https://sass-lang.com/d/slash-div
8:34:02 PM assets.1       |     ╷
8:34:02 PM assets.1       |  16 │     @return ($value / $relative-to) * 1em;
8:34:02 PM assets.1       |     │              ^^^^^^^^^^^^^^^^^^^^^
8:34:02 PM assets.1       |     ╵
8:34:02 PM assets.1       |      node_modules/normalize-scss/sass/normalize/_vertical-rhythm.scss 16:14  normalize-rhythm()
8:34:02 PM assets.1       |      node_modules/normalize-scss/sass/normalize/_vertical-rhythm.scss 27:14  normalize-font-size()
8:34:02 PM assets.1       |      node_modules/normalize-scss/sass/normalize/_normalize-mixin.scss 108:7  normalize()
8:34:02 PM assets.1       |      node_modules/normalize-scss/sass/normalize/_import-now.scss 11:1        @import
8:34:02 PM assets.1       |      src/css/app.scss 14:9                                                   root stylesheet
```

I'm pretty sure there's three solutions:
1. move to [normalize.css](https://necolas.github.io/normalize.css/) - I believe scss can import regular css files no problem and it's most likely to be the same maybe 🤷 
2. move to a different reset css package - I see [sanitize.css](https://github.com/csstools/sanitize.css) and [modern-normalize.css](https://github.com/sindresorhus/modern-normalize)
3. no longer include a css reset package

stephendolan:
I'd personally advocate for option three. Most folks will bring their own CSS framework (I think?), and almost all of them will take care of this aspect anyways.

matthewmcgarvey:
@stephendolan to be honest, it's up to you if no one else chimes in. I'm no CSS expert.

jwoertink:
I like option 2.

[...]

PR #652, opened 2021-07-02
Replace `normalize-scss` with `modern-normalize`

kevinsjoberg:
[normalize-scss](https://github.com/JohnAlbin/normalize-scss) hasn't had an update in three years and its pretty unlikely the following deprecation warning will be fixed any time soon.

Screenshot 2021-07-02 at 21 48 06

[sindresorhus](https://github.com/sindresorhus) maintains [modern-normalize](https://github.com/sindresorhus/modern-normalize) and its regularly kept up-to-date. It's not a 1-1 replacement though, see https://github.com/sindresorhus/modern-normalize#differences-from-normalizecss for details.

Personally, I think it's a fair trade off. If you need something more sophisticated, it's easy enough to swap out.

### DIFF
--- a/src/browser_app_skeleton/package.json
+++ b/src/browser_app_skeleton/package.json
@@ -3,7 +3,7 @@
   "private": true,
   "dependencies": {
     "@rails/ujs": "^6.0.0",
-    "normalize-scss": "^7.0.1",
+    "modern-normalize": "^1.1.0",
     "turbolinks": "^5.2.0"
   },
   "scripts": {
--- a/src/browser_app_skeleton/src/css/app.scss
+++ b/src/browser_app_skeleton/src/css/app.scss
@@ -11,7 +11,7 @@
 // Note: importing with `~` tells webpack to look in the installed npm packages
 // https://stackoverflow.com/questions/39535760/what-does-a-tilde-in-a-css-url-do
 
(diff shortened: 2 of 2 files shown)


matthewmcgarvey:
Connected to https://github.com/luckyframework/lucky_cli/issues/640

You might check out our conversation there. Not sure if it will change what you did at all

kevinsjoberg:
Thanks for that. I did miss the previous conversation. 🙂 I still think it makes sense to ship with something that ensures a consistent experience across the latest version of the most popular browsers. This allow a decent experience out of the box for MVPs and prototypes. Larger projects will most likely have their own opinionated way of working and can easily swap this out.

This of course, is just my take. I'm happy to discuss it further. 🙂

jwoertink (approved):
I don't personally know the differences here, but I'm cool with this change 👍  I agree that shipping with something that just gives it a little "polish" touch is nice. 

Do we need to look at any changes to any styles or markup that's generated from default apps?

jwoertink:
@stephendolan @paulcsmith  I'd like to get your input on this. Both of you handle front-end stuff way more than I.

stephendolan:
This is definitely an improvement. I think we should merge to get rid of the deprecation warning.

Maybe we could remove it altogether in the future, but that can be discussed separately.

PR #653, opened 2021-07-04
Use Lucky::EnforceUnderscoredRoute

paulcsmith:
I think this uses lucky master because of the shard override. Is that right @jwoertink? Seems to be since the specs are not blowing up complaining about a module not existing 😆

### DIFF
--- a/src/web_app_skeleton/src/actions/api_action.cr.ecr
+++ b/src/web_app_skeleton/src/actions/api_action.cr.ecr
@@ -1,6 +1,5 @@
 # Include modules and add methods that are for all API requests
 abstract class ApiAction < Lucky::Action
-  include Lucky::EnforceUnderscoredRoute
   # APIs typically do not need to send cookie/session data.
   # Remove this line if you want to send cookies in the response header.
   disable_cookies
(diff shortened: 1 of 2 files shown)


jwoertink (approved):
Yup! When we have that override file, the generated apps use it as well. https://github.com/luckyframework/lucky_cli/blob/master/.github/workflows/ci.yml#L85

src/web_app_skeleton/src/actions/api_action.cr.ecr
```diff
@@ -12,4 +13,5 @@ abstract class ApiAction < Lucky::Action
   # Add 'include Api::Auth::SkipRequireAuthToken' to your actions to allow all requests.
   include Api::Auth::RequireAuthToken
   <%- end -%>
+  include Lucky::EnforceUnderscoredRoute
```
stephendolan, line 16: @paulcsmith is it intentional that this is included twice in `ApiAction`? It doesn't seem to be inside of an `if` block or anything that would require it.

jwoertink: oh, good call. I missed that... Hmm 🤔  Yeah it seems this last one isn't necessary.

paulcsmith: Yeah it was an accident. Nice catch! Will remove

PR #655, opened 2021-07-06
Replace lucky env module

### LINKED ISSUE #654: Update generated apps config/env to use new LuckyEnv setup
With this PR merged: https://github.com/luckyframework/lucky_env/pull/13 we can now remove the generated `Lucky::Env` module https://github.com/luckyframework/lucky_cli/blob/master/src/web_app_skeleton/config/env.cr

However, we should still keep this file. Instead, this `config/env.cr` file can be loaded up with comments on how to use `LuckyEnv.environment` or any of the different predicate methods like `.development?` `.production?`, etc... as well as the `task?` method. On top of that, we can now use `LuckyEnv.add_env :staging` and such to append additional ones.

kevinsjoberg:
This attempts to replace the `Lucky::Env` module with `LuckyEnv`. I was not entirely sure how we want to go about structuring the `env.cr` file. So any feedback is appreciated. 🙂

Closes #654.

### DIFF
--- a/src/web_app_skeleton/config/env.cr
+++ b/src/web_app_skeleton/config/env.cr
@@ -1,20 +1,33 @@
-module Lucky::Env
-  extend self
+# Environments are managed using `LuckyEnv`. By default, development, production
+# and test are supported. See
+# https://luckyframework.org/guides/getting-started/configuration for details.
+#
+# The default environment is development unless the environment variable
+# LUCKY_ENV is set.
+#
+# Example:
+# ```
+# LuckyEnv.environment  # => "development"
+# LuckyEnv.development? # => true
+# LuckyEnv.production?  # => false
+# LuckyEnv.test?        # => false
+# ```
+#
+# New environments can be added using the `LuckyEnv.add_env` macro.
+#
+# Example:
+# ```
+# LuckyEnv.add_env :staging
+# LuckyEnv.staging? # => false
+# ```
+#
+# To determine whether or not a `LuckyTask` is currently running, you can use
+# the `LuckyEnv.task?` predicate.
+#
+# Example:
+# ```
+# LuckyEnv.task? # => false
+# ```
 
-  {% for env in [:development, :test, :production] %}
-    def {{ env.id }}?
-      name == {{ env.id.stringify }}
-    end
-  {% end %}
-
-  def name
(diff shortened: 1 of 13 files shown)


src/web_app_skeleton/config/env.cr
```diff
@@ -1,20 +1,29 @@
...
+# ```
+#
+# New environments can be added using the `LuckyEnv.add_env` macro.
+#
+# Example:
+# ```
+# LuckyEnv.add_env :staging
+# LuckyEnv.staging? # => false
+# ```
+#
```
jwoertink, line 19: Overall, I really like this. The only thing I think that's missing is mentioning that the value comes from `ENV["LUCKY_ENV"]`..

src/web_app_skeleton/config/env.cr
```diff
@@ -1,20 +1,29 @@
-module Lucky::Env
-  extend self
+# Lucky manages environments using `LuckyEnv`. By default, development,
```
jwoertink, line 1: Can we add a link to the lucky_env github repo?

jwoertink (commented):
Just some small comments on the env.cr config file, but overall this is awesome! I'll look at a new release for the LuckyEnv shard.

src/web_app_skeleton/config/env.cr
kevinsjoberg: I've amended the original commit. See https://github.com/luckyframework/lucky_cli/pull/655/commits/6acd9028931e5f2df70d80d7519e329106b9fd64.

src/web_app_skeleton/config/env.cr
kevinsjoberg: I adjusted the documentation in https://github.com/luckyframework/lucky_cli/pull/655/commits/6acd9028931e5f2df70d80d7519e329106b9fd64. However, instead of linking to the GitHub repository, I link to the configuration section of the Lucky documentation. I thought that would be more helpful. What do you think? 🙂

jwoertink:
I've released v0.1.3 of LuckyEnv. No shard.yml update should be needed. I think it'll just pick it up.

kevinsjoberg:
@jwoertink thank you! 🙂

jwoertink (approved):
:tada:

PR #656, opened 2021-07-09
Fix precompiled task arguments

kevinsjoberg:
Te ensure quoted arguments are handled properly, we need to pass the
array of `String` arguments separately.

As an example, previously the following would not work:

```console
$ lucky exec -e "mate -w"
```


jwoertink (approved):
Ah, nice! Good catch

Issue #597, opened 2021-01-16
Update users migration template to make email column case insensitive


matthewmcgarvey:
With the addition of citext support in https://github.com/luckyframework/avram/pull/608 it seems like a good idea to use it for user emails so that a person who makes an account with "MyAccount@example.com" can sign in with "myaccount@example.com" and no one else can make an account with the same email but with different capitalization.

https://github.com/luckyframework/lucky_cli/blob/00cf9268d654809c7137568d8671be8c07232a08/src/base_authentication_app_skeleton/db/migrations/00000000000001_create_users.cr#L3-L9

We would have to update the migration first to enable the citext extension and then add `case_sensitive: false` to the email line

PR #657, opened 2021-07-11
Make users email column case insensitive

matthewmcgarvey:
Fixes https://github.com/luckyframework/authentic/issues/9
Fixes https://github.com/luckyframework/lucky_cli/issues/597

Before this change, if a user created an account with `Foo@example.com` and then tried to sign in with `foo@example.com` it wouldn't have worked. They could also have made another account with different cases in the email. Now that's not possible.


jwoertink (approved):
🚀

stephendolan (approved):
This is a really great change!

PR #658, opened 2021-07-16
Crystal 1.1.0 support

jwoertink:
Adding Crystal 1.1.0 to the CI. Also changed the `.crystal-version` to be dynamic. It was a little weird that I generated an app with that set to 1.0.0, but the CI that was generated was set to 1.1.0. Now it'll be set to which ever version of Crystal you're using.


matthewmcgarvey (approved):

src/web_app_skeleton/.crystal-version.ecr
```diff
@@ -0,0 +1 @@
+<%= Crystal::VERSION %>
```
kevinsjoberg, line 1: I guess we could end up with a version lower than `1.0.0`? Does that matter?

kevinsjoberg (approved):

src/web_app_skeleton/.crystal-version.ecr
jwoertink: That's a good point. If your local Crystal is set to 0.36.1, and you generate a new project, it'll set that in both this file, and in the CI file. In that case, if you ran in to issues, everyone would say upgrade Crystal first, and then you'd have to just remember/know to update this file. I guess that's also based on if you're even using it. If you installed Crystal through homebrew, then it doesn't matter. This file is only really used by version managers like `asdf`.

The upside to this is that we'd never need to change it again 😂  If I revert it back, and you want to use 1.1.0, you'd have to manually update this file. For a new person, that might be a little wonky.

kevinsjoberg: Yeah, you're right. Thinking about this some more, if you've managed to compile the Lucky CLI, I'd expect it to work, given we set constraints on the Crystal version. 🙂

PR #661, opened 2021-07-18
Update the seed docs to use an existing operation.

### LINKED ISSUE #638: SampleData task comment uses non-existant save operation
https://github.com/luckyframework/lucky_cli/blob/98d5bf099723a28b40c9052d4b534c1ecc10afc7/src/web_app_skeleton/tasks/db/seed/sample_data.cr#L18

Makes sense to update this comment to use the operation we already provide (even if the user chose not to include authentication).

```crystall
SignUpUser.create!(email: "admin@example.com", password: "test123", password_confirmation: "test123")
```

jwoertink:
Fixes #638

### DIFF
--- a/src/web_app_skeleton/tasks/db/seed/sample_data.cr
+++ b/src/web_app_skeleton/tasks/db/seed/sample_data.cr
@@ -21,8 +21,9 @@ class Db::Seed::SampleData < LuckyTask::Task
     # You likely want to be able to run this file more than once. To do that,
     # only create the record if it doesn't exist yet:
     # ```
-    # # be sure to set `upsert_lookup_columns :email` in the operation
-    # SignUpUser.upsert!(email: "me@example.com", password: "test123", password_confirmation: "test123")
+    # if UserQuery.new.email("me@example.com").none?
+    #   SignUpUser.create!(email: "me@example.com", password: "test123", password_confirmation: "test123")
+    # end
     # ```
     puts "Done adding sample data"
   end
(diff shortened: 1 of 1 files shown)


kevinsjoberg (approved):
Nice. 🎉

src/web_app_skeleton/tasks/db/seed/sample_data.cr
```diff
@@ -8,21 +8,22 @@ class Db::Seed::SampleData < LuckyTask::Task
...
     #
     # You likely want to be able to run this file more than once. To do that,
     # only create the record if it doesn't exist yet:
-    #
-    # unless UserQuery.new.email("me@example.com").first?
-    #  SaveUser.create!(email: "me@example.com", name: "Jane")
-    # end
+    # ```
+    # # be sure to set `upsert_lookup_columns :email` in the operation
+    # SignUpUser.upsert!(email: "me@example.com", password: "test123", password_confirmation: "test123")
```
matthewmcgarvey, line 25: This doesn't do the same thing that the comment says. Upsert will update, but the comment is referring to a find or create.

jwoertink: I was trying to be clever 😛  I guess I could change the example to do `if UserQuery.new.email().none?` and then use the normal Signupuser.create?

matthewmcgarvey: Yea, that's what I was thinking

matthewmcgarvey (approved):

PR #666, opened 2021-07-22
Fixing this spec that should never have worked

jwoertink:
but for some reason it passed somethings... Also added a few extra specs that were missing anyway

### DIFF
--- a/spec/integration/init_web_spec.cr
+++ b/spec/integration/init_web_spec.cr
@@ -28,7 +28,7 @@ describe "Initializing a new web project" do
   it "creates a full web app with generator" do
     puts "Web app generators: Running integration spec. This might take awhile...".colorize(:yellow)
     with_project_cleanup do
-      should_run_successfully "crystal run src/lucky.cr -- init.custom test-project"
+      should_run_successfully "crystal run src/lucky.cr -- init.custom test-project --no-auth"
 
       FileUtils.cd "test-project" do
         should_run_successfully "shards install --ignore-crystal-version"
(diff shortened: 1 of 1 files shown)


spec/integration/init_web_spec.cr
```diff
@@ -28,7 +28,7 @@ describe "Initializing a new web project" do
   it "creates a full web app with generator" do
     puts "Web app generators: Running integration spec. This might take awhile...".colorize(:yellow)
     with_project_cleanup do
-      should_run_successfully "crystal run src/lucky.cr -- init.custom test-project"
+      should_run_successfully "crystal run src/lucky.cr -- init.custom test-project --no-auth"
```
jwoertink, line 31: We test `gen.model User` which always fails because we were generating auth. Alternatively, we could keep generating auth, but test generating a different model.

spec/integration/init_web_spec.cr
```diff
@@ -39,14 +39,22 @@ describe "Initializing a new web project" do
         should_run_successfully "lucky gen.page Users::IndexPage"
         should_run_successfully "lucky gen.component Users::Header"
         should_run_successfully "lucky gen.resource.browser Comment title:String"
+        should_run_successfully "lucky gen.task email.monthly_update"
+        should_run_successfully "lucky gen.secret_key"
```
jwoertink, line 43: We didn't test these before, but they should be here.

spec/integration/init_web_spec.cr
```diff
@@ -39,14 +39,22 @@ describe "Initializing a new web project" do
...
         File.read("src/actions/comments/index.cr").should contain "Comments::Index"
+        File.read("src/pages/comments/index_page.cr").should contain "Comments::IndexPage"
+        File.read("src/models/comment.cr").should contain "Comment < BaseModel"
+        File.read("src/queries/comment_query.cr").should contain "CommentQuery"
+        File.read("src/operations/save_comment.cr").should contain "SaveComment"
         File.read("src/actions/api/users/show.cr").should_not be_nil
         File.read("src/actions/users/show.cr").should_not be_nil
         File.read("src/pages/users/index_page.cr").should_not be_nil
         File.read("src/components/users/header.cr").should contain "Users::Header < BaseComponent"
-        should_run_successfully "crystal build src/start_server.cr"
```
jwoertink, line 48: We said build this file, but never actually ran the setup. If you booted this app before, it would have failed with pending migrations and such.

spec/integration/init_web_spec.cr
```diff
@@ -28,7 +28,7 @@ describe "Initializing a new web project" do
   it "creates a full web app with generator" do
     puts "Web app generators: Running integration spec. This might take awhile...".colorize(:yellow)
     with_project_cleanup do
```
jwoertink, line 30: This line drops the DB after it's done. However, since we never ran the setup in this spec, no DB was ever created. This should have always failed since it was added in November 2020, but if you look at the last few merged PRs, you'll see it's been passing....  (👻 )

jwoertink:
Here's what's causing the fail. This is due to https://github.com/luckyframework/lucky/pull/1551 which is fixed for the next release. 
Screen Shot 2021-07-22 at 2 46 20 PM

PR #679, opened 2021-09-20
explicitly sends the manifest file name in generated app

robacarp:
This change isn't _necessary_ but it might reduce the number of "hey how can I change this" help requests.


jwoertink (approved):
That's good enough for me! :+1:  Thanks

jwoertink (changes requested):
ooo.... it broke :thinking:  Maybe because it needs to be linked to Lucky master?

robacarp:
Yep that did it.

jwoertink (approved):

PR #683, opened 2021-09-21
Updated luckycli that points to the luckycli website for installation

### LINKED ISSUE #680: Remove installing docs from README
The README file in this repo has instructions on how to install the CLI. It's technically wrong because it has you setup the CLI from the master branch which may not always be "working". If you're using Linux, chances are, you might look here first for installing it.

Instead, we should point to the website guides so it says something like "To install the Lucky CLI, read the guides for your Operating System https://luckyframework.org/guides/getting-started/installing".

Then the remainder of the README should be explicitly for how to install the CLI for the purposes of contributing to the project itself (i.e. submitting a PR).

Ayush0431:
Fixes #680

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,7 @@ A Crystal library for creating and running tasks
 
 ## Installing the CLI by Homebrew
 
-1.  Install [homebrew](http://brew.sh)
-2.  Run `brew tap luckyframework/lucky`
-3.  Run `brew install lucky`
+To install the Lucky CLI, read the guides for your Operating System https://luckyframework.org/guides/getting-started/installing
 
 ## Building CLI from source
 
--- a/README.md
+++ b/README.md
@@ -2,12 +2,13 @@
 
 A Crystal library for creating and running tasks
 
-## Installing the CLI by Homebrew
+##Installing the CLI by Homebrew
 
-To install the Lucky CLI, read the guides for your Operating System https://luckyframework.org/guides/getting-started/installing
+To install the Lucky CLI, read the guides for your Operating System [here](https://luckyframework.org/guides/getting-started/installing)
 
-## Building CLI from source
+##Development
 
+**Building the CLI
 1.  Install [crystal](https://github.com/crystal-lang/crystal)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,9 +4,7 @@ A Crystal library for creating and running tasks
 
 ## Installing the CLI by Homebrew
 
-1.  Install [homebrew](http://brew.sh)
-2.  Run `brew tap luckyframework/lucky`
-3.  Run `brew install lucky`
+To install the Lucky CLI, read the guides for your Operating System https://luckyframework.org/guides/getting-started/installing
```
jwoertink, line 7: This will need to replace the building from source that's below. I'm thinking both of the sections go away, and it's replaced by a single block that says "Installing Lucky". Then the URL should be a clickable link.

Then the next section below would saying something along the lines of "Development" with a sub sections of "building the CLI. That subsection could then detail out the steps to build from source.

jwoertink (changes requested):
Hey @Ayush0431! I appreciate you taking this on. Welcome to the community :heart: 

I think this is going to need a lot more than just a single line change. The document currently shows 2 different ways to install the CLI depending on your OS, but both of those sections can be consolidated in to a single section that links to the site. 

I left a note on there, but please let me know if it's unclear. It may require a bit of extra thought in to the wording. It doesn't need to say exactly what I've said, but just get the proper point across using my suggestions as a template.

README.md
Ayush0431: hey @jwoertink ,I got your point ,I'll make the required changes according to your suggestions.

Ayush0431: hey @jwoertink I have done changes according to your suggestions can u please check it . 
can u review this once https://github.com/luckyframework/lucky_cli/pull/683/commits/e6bcc50b5f0acb7c09907370abdd2928c66618ca

jwoertink (approved):
Thanks for making the changes. It's still not really what I'm looking for, but I think maybe I wasn't very clear on what needed to be done.

Since this is just the README, what I'll do is I can just merge this in as is, then I'll adjust the other things after.

I appreciate you taking this on and helping out!

PR #706, opened 2021-12-08
Seed task uses log instead of puts

wontruefree:
This moves to use puts so you can suppress output if you want.


jwoertink (commented):
I like the idea of being able to suppress the output, but if it's just `Log`, then you'd have to turn that off globally which might not be wanted. We could look at doing something like `LuckyTask::Log.info`, then you can tune it for tasks specifically... But really, I think maybe this should actually be `output.puts` and use the `output` property on LuckyTask

https://github.com/luckyframework/lucky_task/blob/da2bfd5d69c6aab7e5c4a1e998469645d4b31a63/src/lucky_task/task.cr#L6

Then maybe we add a habitat config for tasks that allow you to just turn off all output. What do you think about going with one of those approaches?

PR #707, opened 2021-12-10
Cleaning up CI

jwoertink:
We need to be able to catch any nightly changes in Crystal against the latest Lucky. Also testing that edge Lucky works on the latest Crystal.

Cleaned up the check_format since we don't really need to test that it formats on all the different Crystal versions. As long as it's clean on the latest, that should be fine enough.

### DIFF
--- a/.github/workflows/weekly.yml
+++ b/.github/workflows/weekly.yml
@@ -1,49 +1,24 @@
-name: Lucky CLI CI
+name: Lucky CLI Weekly CI
 
 on:
   push:
     branches: [master]
   pull_request:
-    branches: "*"
+  schedule:
+    - cron: "0 1 * * MON"
+  workflow_dispatch:
 
 jobs:
-  check_format:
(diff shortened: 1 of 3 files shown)


matthewmcgarvey (approved):
We might also consider a periodic run of the edge one since Lucky will be updated more than here

jwoertink:
I like that idea. Any clue how I can set the cron to only run on a specific version?

jwoertink:
ok, I think just having the second workflow will work

.github/workflows/weekly.yml
```diff
@@ -0,0 +1,58 @@
+name: Lucky CLI Weekly CI
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+  schedule:
+    - cron: "0 1 * * MON"
```
jwoertink, line 8: Run every monday at 1am

PR #710, opened 2021-12-24
Add nox as a ProcessRunner

matthewmcgarvey:
[nox](https://github.com/matthewmcgarvey/nox) is a process manager that I have crated written completely in Crystal. The motivation for this can be found in a recent wishlist discussion made by @jwoertink [here](https://github.com/luckyframework/lucky/discussions/1629).

### DIFF
--- a/spec/lucky_cli/process_runner_spec.cr
+++ b/spec/lucky_cli/process_runner_spec.cr
@@ -54,12 +54,4 @@ describe LuckyCli::ProcessRunner do
     InstalledRunner.new.installed?.should be_true
     MissingRunner.new.installed?.should be_false
   end
-
-  it "can find an installed process runner and start it" do
-    process = FakeProcess.new
-
-    LuckyCli::ProcessRunner.start(process)
-
-    process.command.not_nil!.empty?.should be_false
-  end
 end
--- a/src/lucky_cli/process_runner.cr
+++ b/src/lucky_cli/process_runner.cr
@@ -42,6 +42,24 @@ module LuckyCli
     end
   end
 
+  class ProcessRunner::Nox < ProcessRunner
+    def installed? : Bool
+      true
+    end
+
+    def start(process = Process)
+      ::Nox.run("Procfile.dev")
+    end
+
+    def command
+      "DOES NOTHING"
+    end
+
+    def start_args
+      # does nothing
+    end
+  end
+
   class ProcessRunner::Overmind < ProcessRunner
     def command
       "overmind"
--- a/shard.yml
+++ b/shard.yml
@@ -19,6 +19,9 @@ dependencies:
   lucky_task:
     github: luckyframework/lucky_task
     version: ~> 0.1.1
+  nox:
+    github: matthewmcgarvey/nox
+    branch: main
 
 development_dependencies:
   ameba:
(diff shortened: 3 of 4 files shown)


src/lucky_cli/process_runner.cr
```diff
@@ -42,6 +42,24 @@ module LuckyCli
     end
   end
 
+  class ProcessRunner::Nox < ProcessRunner
```
matthewmcgarvey, line 45: For now I did the minimal change just to add in Nox to the mix. I will follow up this PR with another that rips out all this ProcessRunner stuff since we will no longer need it. 

The resulting code will literally just be:

```crystal
class LuckyCli::Dev < LuckyTask::Task
  summary "Start your app with a process runner and Procfile.dev"

  def call
    Nox.run("Procfile.dev")
  end
end
```

src/lucky_cli/process_runner.cr
```diff
@@ -42,6 +42,24 @@ module LuckyCli
...
+    def installed? : Bool
+      true
```
matthewmcgarvey, line 47: True since it a dependency included with the shard

src/lucky_cli/process_runner.cr
```diff
@@ -42,6 +42,24 @@ module LuckyCli
...
+    end
+
+    def start(process = Process)
+      ::Nox.run("Procfile.dev")
+    end
+
+    def command
+      "DOES NOTHING"
```
matthewmcgarvey, line 55: Have to return a string here

spec/lucky_cli/process_runner_spec.cr
```diff
@@ -54,12 +54,4 @@ describe LuckyCli::ProcessRunner do
     InstalledRunner.new.installed?.should be_true
     MissingRunner.new.installed?.should be_false
   end
-
-  it "can find an installed process runner and start it" do
```
matthewmcgarvey, line 58: Removed this spec since it was using the actually ProcessRunners and since Nox is the first one, it was using that. The nox runner doesn't use the FakeProcess so it would have hung indefinitely.

jwoertink (commented):
I like the changes, but I want to try and get this in to a project today so I can play with it directly before making the final approval.

jwoertink (approved):
Sweet! It works great. There's some display issues with coloring and such, but I'll open an issue on nox. It may be irrelevant

stephendolan:
Should we make sure the box test suite has tests before making this an official change, or is it difficult to test a process runner?

jwoertink:
What do you mean? it literally say "it works" 😂 

```
it "works" do
    false.should eq(true)
  end
```

see?!

matthewmcgarvey:
I was largely referencing https://github.com/DarthSim/hivemind to make Nox and it has no tests. It's not really an excuse so much as a... well it's kind of an excuse 🙈 

Let me see what I can reasonably test

stephendolan:
I tested a lot of command line output with Spark, if it's helpful. No idea how "good" this approach is, though:

https://github.com/stephendolan/spark

matthewmcgarvey:
I have added specs to nox along with a CI process. Everybody good with merging?

stephendolan:
I haven't used it, but as long as it simplifies getting up and running and doesn't require any manual steps to add/support for Lucky users, I'm in favor!

jwoertink:
Yup. Looks good to me!

PR #711, opened 2021-12-24
Use new separated lucky and avram

matthewmcgarvey:
Updates the cli to incorporate the changes of https://github.com/luckyframework/lucky/pull/1620 and https://github.com/luckyframework/avram/pull/772.

### DIFF
--- a/src/web_app_skeleton/src/shards.cr.ecr
+++ b/src/web_app_skeleton/src/shards.cr.ecr
@@ -3,8 +3,9 @@ require "lucky_env"
 LuckyEnv.load?(".env")
 
 # Require your shards here
-require "avram"
 require "lucky"
+require "avram"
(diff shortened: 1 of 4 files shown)


src/web_app_skeleton/src/shards.cr.ecr
```diff
@@ -3,8 +3,9 @@ require "lucky_env"
 LuckyEnv.load?(".env")
 
 # Require your shards here
-require "avram"
 require "lucky"
+require "avram"
+require "avram/lucky"
```
robacarp, line 8: Could you make "avram/lucky" require the base avram library?

matthewmcgarvey: It already does, but I went the more explicit route

PR #719, opened 2021-12-30
Update GitHub workflow Postgre host to localhost

### LINKED ISSUE #718: Update the CI DB HOST to localhost
https://github.com/luckyframework/lucky_cli/blob/91d4149364b177bb1071e26847ff0431ed658b17/src/web_app_skeleton/.github/workflows/ci.yml.ecr#L40

I'm pretty sure this line can just go away... Or it may have to be set to `localhost`... Basically, we used to run on actual Crystal containers and Github auto mapped `postgres` as the host to the service. Now we just run directly on the `ubuntu-latest` machine, so the host is just `localhost`. 

Avram can be used as an example here too if need be.
https://github.com/luckyframework/avram/blob/9b184d4b586d783b096a5d377eafd98c78e32026/.github/workflows/ci.yml#L82

robcole:
Resolves a bug related to recent changes to run tests
directly on Ubuntu instead of a Docker container.

Here's a smoke test I ran with it: https://github.com/robcole/lucky-cli-postgre-workflow-fix/runs/4664076513?check_suite_focus=true

Resolves #718

### DIFF
--- a/src/web_app_skeleton/.github/workflows/ci.yml.ecr
+++ b/src/web_app_skeleton/.github/workflows/ci.yml.ecr
@@ -37,7 +37,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       LUCKY_ENV: test
-      DB_HOST: postgres
+      DB_HOST: localhost
     continue-on-error: ${{ matrix.experimental }}
     services:
       postgres:


matthewmcgarvey (approved):

jwoertink:
Thanks @robcole!

PR #720, opened 2022-01-01
Remove ProcessRunner code leaving Nox to run Procfiles

### LINKED ISSUE #688: Using Heroku as process manager on M1 mac seems to have issues
If you're on an M1 Mac, and using Heroku for your local process manager, you might run `lucky dev` and things blow up. 

```
execution of command failed with code: 1: cc "${@}" -o /Users/user/.cache/crystal/Users-user-Repositories-robacarp-offline_pink-lib-lucky-src-run_macros-generate_asset_helpers.cr/macro_run  -rdynamic -L/opt/homebrew/lib -L/opt/crystal-ci-univ2nd-dev-1/embedded/lib -lpcre -lgc -lpthread -L/opt/homebrew/Cellar/libevent/2.1.12/lib -levent -liconv -ldl
```

For now, install Overmind to get around it until we can find a proper solution

### LINKED ISSUE #714: Generated app should check if foreman is installed
In the generated app, it should check if foreman is installed in https://github.com/luckyframework/lucky_cli/blob/master/src/web_app_skeleton/script/system_check.ecr

If foreman isn't installed, `lucky dev` will fail to run

matthewmcgarvey:
Closes #714 
Closes #688

The [previous PR](https://github.com/luckyframework/lucky_cli/pull/710) added [Nox](https://github.com/matthewmcgarvey/nox) as a process runner. Since it's a shard instead of a process installed on users' machines, it's guaranteed to be present so we don't need the `ProcessRunner` code anymore. Consequently also removed the use of the `ensure_process_runner_installed` task which I never even knew existed.

This is a breaking change. Projects will need to update `script/system_check` and delete these lines

```
# Only if this isn't CI
if [ -z "$CI" ]; then
  lucky ensure_process_runner_installed
fi
```


jwoertink (approved):
Sweet!

PR #725, opened 2022-01-11
Adding Crystal 1.3.0 to CI

jwoertink:
use only the min and max crystal versions for the CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
           - shard.override.yml
         crystal_version:
           - 1.0.0
-          - 1.3.0
+          - latest
         experimental:
           - false
         include:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -33,8 +33,7 @@ jobs:
           - shard.override.yml
         crystal_version:
           - 1.0.0
-          - 1.1.0
-          - 1.2.0
+          - 1.3.0
```
matthewmcgarvey, line 36:
```suggestion
          - latest
```

jwoertink: I thought the same thing.... I'm wondering though, would this ever be an issue where maybe a Crystal version is released, and this breaks something? Now thinking about it, once 2.0 drops, we will have to change all of this to not be 2.0 until we know for sure Lucky is supported. 🤔  I guess that's a problem for a later day

matthewmcgarvey: Since we're also running specs on nightly, we would know about it ahead of time

matthewmcgarvey (approved):

PR #726, opened 2022-01-15
Replacing Turbolinks with @hotwired/turbo

rmarronnier:
Sister PR of https://github.com/luckyframework/lucky/pull/1648

### DIFF
--- a/src/browser_app_skeleton/package.json
+++ b/src/browser_app_skeleton/package.json
@@ -4,7 +4,7 @@
   "dependencies": {
     "@rails/ujs": "^6.0.0",
     "modern-normalize": "^1.1.0",
-    "turbolinks": "^5.2.0"
+    "@hotwired/turbo": "^7.1.0"
   },
   "scripts": {
     "heroku-postbuild": "yarn prod",
--- a/src/browser_app_skeleton/src/js/app.js
+++ b/src/browser_app_skeleton/src/js/app.js
@@ -4,11 +4,11 @@
 // Though it says "Rails" it actually works with any framework.
 require("@rails/ujs").start();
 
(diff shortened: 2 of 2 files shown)


jwoertink:
Not sure why all of the CI is failing, but I've added this in to the website https://github.com/luckyframework/website/pull/932 We can test on a live site that it's fine before making the final merge.

jwoertink (changes requested):
We will need the HTML head tags updated as well

https://github.com/luckyframework/lucky_cli/blob/5649c9e78c9c359710347ef0e91fa87c04072ea6/src/browser_app_skeleton/src/components/shared/layout_head.cr#L8-L10

rmarronnier:
Yes, the CI is failing and because the global specs using the support `should_run_successfully` method don't return  the "downstream" error text, I'm blind. (Problem A)

> We can test on a live site that it's fine before making the final merge.

Maybe it's enough, but I wanted to create specific lucky flowspecs to check we have the same behavior between before (@rails/ujs + turbolinks) and after (@hotwired/turbo only). (Problem B)

Right now, only the `browser_authentification_app_skeleton` uses lucky_flow, and for specific auth logic.

I have to tackle problem A before working on problem B, or I'll tiptoe in the dark, wondering why the specs failed.

Also for reference :

jwoertink:
Looks like this just became a bit larger of a PR:
A comparable version of this handler would need to be created:
https://github.com/luckyframework/lucky/blob/main/src/lucky/redirectable_turbolinks_support.cr

Then whatever needs to be done to ensure this stuff works https://github.com/luckyframework/website/issues/947

Just dropping these links here so we have a spot to reference them.

PR #727, opened 2022-01-16
Update default watcher port to 3000

### LINKED ISSUE #704: Port of generated lucky apps (5000) conflicts with latest Mac release Control Center
On the latest release of macOS Monterey, the Control Center app uses port 5000. I remember hearing about this, but I just spent 10 minutes trying to figure out why my lucky app would work for a bit and then I would start getting 403 Forbidden responses. This is not something we can reasonably expect developers to know about, so I think it might be best to change it to something else.

It is defined here https://github.com/luckyframework/lucky_cli/blob/6d247a4b92cfbf49364128f93cf29164d468e528/src/web_app_skeleton/config/watch.yml#L2

Might be worth noting that browsersync has it's own port of 5001, I believe, and doesn't have problems.

robcole:
Port 5000 is used by Apple internal services. Rails has
been defaulting to port 3000 for years and hasn't run
into any issues, so this seems like a very safe bet.

Resolves #704


jwoertink (approved):
Thanks for doing this!

PR #734, opened 2022-02-28
poking at a docker compose workflow

robacarp:
I think, based on some chatter in discord, that this will be helpful for some newcomers. 

It seems like a _bad_ idea to require that people use docker compose, but it seems like a _great_ idea to provide people easier routes to get started if they're into docker compose workflows. My hope is to enable that without creating too much clutter, and without altering a default workflow.

I know everyone has their own style of docker stuff, and that a bunch of community tools have expectations about how things are done too. We'll have to decide as a group which styles and standards to embrace. Comments welcome!

### DIFF
--- a/src/web_app_skeleton/docker-compose.yml
+++ b/src/web_app_skeleton/docker-compose.yml
@@ -31,6 +31,10 @@ services:
         source: postgres_data
         target: /var/lib/postgresql
     ports:
+      # The postgres database container is exposed on the host at port 6543 to
+      # allow connecting directly to it with postgres clients. The port differs
+      # from the postgres default to avoid conflict with existing postgres
+      # servers.
       - 6543:5432
 
 volumes:
--- a/src/web_app_skeleton/docker-compose.yml
+++ b/src/web_app_skeleton/docker-compose.yml
@@ -1,0 +1,39 @@
+version: '3.8'
+services:
+  lucky:
+    build:
+      context: .
+      dockerfile: docker/development.dockerfile
+    environment:
+      DATABASE_URL: postgres://postgres@postgres:5432/postgres
+    volumes:
+      - type: bind
+        source: .
+        target: /app
+      - type: volume
+        source: node_modules
+        target: /app/node_modules
+      - type: volume
+        source: shards_lib
+        target: /app/lib
+    depends_on:
+      - postgres
+
+    entrypoint: ["docker/dev_entrypoint.sh"]
+
+  postgres:
+    image: postgres:14-alpine
+    environment:
+      POSTGRES_USER: lucky
+      POSTGRES_PASSWORD: developer
+    volumes:
+      - type: volume
+        source: postgres_data
+        target: /var/lib/postgresql
+    ports:
+      - 6543:5432
+
+volumes:
+  postgres_data:
+  node_modules:
+  shards_lib:
--- a/src/web_app_skeleton/docker/dev_entrypoint.sh
+++ b/src/web_app_skeleton/docker/dev_entrypoint.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# This is the entrypoint script used for development docker workflows. By
+# default it will:
+#  - Install shards.
+#  - Run migrations.
+#  - Start the dev server.
+# It also accepts any commands to be run instead.
+
+
+warnfail () {
+  echo "$@" >&2
+  exit 1
+}
+
+case ${1:-} in
+  "") # If no arguments are provided, start lucky dev server.
+    ;;
+
+  *) # If any arguments are provided, execute them instead.
+    exec "$@"
+esac
+
+echo "Installing npm packages..."
+npm install
+
+if ! shards check ; then
+  echo "Installing shards..."
+  shards install
+fi
+
+echo "Waiting for postgres to be available..."
+./docker/wait-for-it.sh postgres:5432
+
+echo "Starting lucky dev server..."
+exec lucky dev || warnfail "Lucky dev server failed."
--- a/src/web_app_skeleton/docker/dev_entrypoint.sh
+++ b/src/web_app_skeleton/docker/dev_entrypoint.sh
@@ -35,4 +35,4 @@ echo "Waiting for postgres to be available..."
 ./docker/wait-for-it.sh postgres:5432
 
 echo "Starting lucky dev server..."
-exec lucky dev || warnfail "Lucky dev server failed."
+exec lucky dev
--- a/src/web_app_skeleton/docker/development.dockerfile
+++ b/src/web_app_skeleton/docker/development.dockerfile
@@ -1,0 +1,31 @@
+FROM crystallang/crystal:1.1.1
+
+# Add the nodesource ppa to apt.
+RUN wget https://deb.nodesource.com/setup_16.x -O- | bash
+
+# Install nodejs, npm, yarn
+RUN apt-get update
+RUN apt-get install -y nodejs
+RUN npm install -g yarn mix
+
+# Install postgres cli tools
+RUN apt-get install -y postgresql-client
+
+# Installs overmind, not needed if nox is the process manager.
+RUN apt-get install -y tmux && \
+    wget https://github.com/DarthSim/overmind/releases/download/v2.2.2/overmind-v2.2.2-linux-amd64.gz && \
+    gunzip overmind-v2.2.2-linux-amd64.gz && \
+    mv overmind-v2.2.2-linux-amd64 /usr/bin/overmind && \
+    chmod +x /usr/bin/overmind
+
+# Install lucky cli, TODO: fetch current lucky version from source code.
+WORKDIR /lucky/cli
+RUN git clone https://github.com/luckyframework/lucky_cli . && \
+    git checkout v0.29.0 && \
+    shards build --without-development && \
+    cp bin/lucky /usr/bin
+
+WORKDIR /app
+ENV DATABASE_URL=postgres://postgres:postgres@host.docker.internal:5432/postgres
+EXPOSE 5001
+
--- a/src/web_app_skeleton/docker/wait-for-it.sh
+++ b/src/web_app_skeleton/docker/wait-for-it.sh
@@ -1,0 +1,189 @@
+#!/usr/bin/bash
+#
+# Pulled from https://github.com/vishnubob/wait-for-it on 2022-02-28.
+# Licensed under the MIT license as of 81b1373f.
+#
+# Below this line, wait-for-it is the original work of the author.
+#
+# Use this script to test if a given TCP host/port are available
+
+WAITFORIT_cmdname=${0##*/}
+
+echoerr() { if [[ $WAITFORIT_QUIET -ne 1 ]]; then echo "$@" 1>&2; fi }
+
+usage()
+{
+    cat << USAGE >&2
+Usage:
+    $WAITFORIT_cmdname host:port [-s] [-t timeout] [-- command args]
+    -h HOST | --host=HOST       Host or IP under test
+    -p PORT | --port=PORT       TCP port under test
+                                Alternatively, you specify the host and port as host:port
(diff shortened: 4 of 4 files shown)


src/web_app_skeleton/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+      - type: volume
+        source: node_modules
+        target: /app/node_modules
+      - type: volume
+        source: shards_lib
+        target: /app/lib
+    depends_on:
+      - postgres
+
+    entrypoint: ["docker/dev_entrypoint.sh"]
```
robacarp, line 22: I opted to embed the entrypoint in a script rather than simply inline the startup logic to keep this file terse. In the long run I think the entrypoint script will likely grow to be more helpful and I don't think it'll live inline here reasonably very well.

src/web_app_skeleton/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+    image: postgres:14-alpine
+    environment:
+      POSTGRES_USER: lucky
+      POSTGRES_PASSWORD: developer
+    volumes:
+      - type: volume
+        source: postgres_data
+        target: /var/lib/postgresql
+    ports:
+      - 6543:5432
```
robacarp, line 34: The postgres port is forwarded to the host so that someone can connect to it easily with native gui/cli tools. I chose to expose it on a port different from the typical postgres port to avoid conflict if someone already has postgres up and running on their workstation.

src/web_app_skeleton/docker/dev_entrypoint.sh
```diff
@@ -0,0 +1,38 @@
...
+esac
+
+echo "Installing npm packages..."
+npm install
+
+if ! shards check ; then
+  echo "Installing shards..."
+  shards install
+fi
+
```
robacarp, line 33: I played around with running migrations automatically, but ultimately decided it wasn't worth it. There are a bunch of reasons why I can think that would be a bad idea, and it _really_ slows down startup because migrations require compilation before they can even be checked.

src/web_app_skeleton/docker/development.dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+# Install postgres cli tools
+RUN apt-get install -y postgresql-client
+
+# Installs overmind, not needed if nox is the process manager.
+RUN apt-get install -y tmux && \
+    wget https://github.com/DarthSim/overmind/releases/download/v2.2.2/overmind-v2.2.2-linux-amd64.gz && \
+    gunzip overmind-v2.2.2-linux-amd64.gz && \
+    mv overmind-v2.2.2-linux-amd64 /usr/bin/overmind && \
+    chmod +x /usr/bin/overmind
```
robacarp, line 19: I think this is probably not necessary with @matthewmcgarvey's nox stuff, but I'm working off of the current release tag so I found it necessary. 

TODO: ❌

src/web_app_skeleton/docker/development.dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+# Install lucky cli, TODO: fetch current lucky version from source code.
+WORKDIR /lucky/cli
+RUN git clone https://github.com/luckyframework/lucky_cli . && \
+    git checkout v0.29.0 && \
+    shards build --without-development && \
+    cp bin/lucky /usr/bin
```
robacarp, line 26: I know I've seen some code to open the shard.yml and extract the version of a tool, but I couldn't find it. I imagine there's a `crystal --eval` line I can drop in here to automatically install the right version.

src/web_app_skeleton/docker/development.dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+WORKDIR /app
+ENV DATABASE_URL=postgres://postgres:postgres@host.docker.internal:5432/postgres
```
robacarp, line 29: This line isn't necessary but _if_ someone wants to boot the docker container directly rather than via compose, this line serves as a guidepost for how they'll need to connect it to the postgres instance on their local mac. I'm not sure if it's necessary to keep or not.

robacarp:
There are some drawbacks to this setup, but I don't think they are blockers. It may be better to avoid `lucky dev` and instead run the node and crystal processes directly in their own containers. That would greatly decrease the complexity of the dockerfile and it would likely speed up startup a ton because docker's parallelization of compose containers.

src/web_app_skeleton/docker/development.dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM crystallang/crystal:1.1.1
```
grepsedawk, line 1: Might as well use latest here imho

robacarp: Unfortunately, on M1, I ran into [some issues](https://github.com/crystal-lang/crystal/issues/11703) -- for now this is just 1.1.1 so I could get it moving forward but yes this should be updated to be correct before merge.

src/web_app_skeleton/docker/dev_entrypoint.sh
grepsedawk: Line 38 could maybe be:
`lucky dev || lucky db.migrate || warnfail "Lucky dev server failed."`

src/web_app_skeleton/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+        target: /app/node_modules
+      - type: volume
+        source: shards_lib
+        target: /app/lib
+    depends_on:
+      - postgres
+
+    entrypoint: ["docker/dev_entrypoint.sh"]
+
+  postgres:
```
grepsedawk, line 24: One thing we could include for all is to boot this postgres instance (and auto-connect) in the normal `Procfile`, since then we can make lucky auto connect to dev DB's

src/web_app_skeleton/docker/dev_entrypoint.sh
robacarp: That could produce some weird experiences. If lucky dev runs for 20 min and then fails because of a crystal compiler bug, the app would try to run migrations.

src/web_app_skeleton/docker/dev_entrypoint.sh
```diff
@@ -0,0 +1,38 @@
...
+echo "Waiting for postgres to be available..."
+./docker/wait-for-it.sh postgres:5432
+
+echo "Starting lucky dev server..."
+exec lucky dev || warnfail "Lucky dev server failed."
```
refi64, line 38: The warning will never be printed unless `lucky` is not found due to the `exec`.

src/web_app_skeleton/docker-compose.yml
robacarp: I'm not sure what you mean.

src/web_app_skeleton/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+    build:
+      context: .
+      dockerfile: docker/development.dockerfile
+    environment:
+      DATABASE_URL: postgres://postgres@postgres:5432/postgres
+    volumes:
+      - type: bind
+        source: .
+        target: /app
+      - type: volume
```
refi64, line 13: it's not clear to me why these have to be separate modules? Kind of ties into what I mentioned about build steps typically being performed at image build time I guess

src/web_app_skeleton/docker/dev_entrypoint.sh
```diff
@@ -0,0 +1,38 @@
...
+  *) # If any arguments are provided, execute them instead.
+    exec "$@"
+esac
+
+echo "Installing npm packages..."
+npm install
+
+if ! shards check ; then
+  echo "Installing shards..."
+  shards install
```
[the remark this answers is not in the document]
refi64: (this does also mean you don't need your build tools or `node_modules` in the docker image, making them MUCH smaller!)

src/web_app_skeleton/docker-compose.yml
refi64: Usually these are in the Dockerfile, so the same image could be run standalone without a rebuild

src/web_app_skeleton/docker-compose.yml
refi64: (that info might be useful in a comment here!)

src/web_app_skeleton/docker/dev_entrypoint.sh
grepsedawk: This isn't hosting, this is the dev entrypoint

src/web_app_skeleton/docker-compose.yml
robacarp: I started with it there, actually. I removed it because this dockerfile is intended to be used with the compose, and the file doesn't exist without the compose mounted volume. If the entrypoint were coded into the docker file, the file would also have to be copied into the container as well.

src/web_app_skeleton/docker/development.dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+# Add the nodesource ppa to apt.
+RUN wget https://deb.nodesource.com/setup_16.x -O- | bash
+
+# Install nodejs, npm, yarn
+RUN apt-get update
+RUN apt-get install -y nodejs
+RUN npm install -g yarn mix
```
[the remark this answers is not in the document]
robacarp: Let me know what you think about this: https://stackoverflow.com/a/47451019/497719

I did combine the steps of some sequences because they're unlikely to be used independently. But for these specific examples it seemed like it would hamper development workflows to combine them.

src/web_app_skeleton/docker/development.dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+# Install postgres cli tools
+RUN apt-get install -y postgresql-client
```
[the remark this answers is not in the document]
robacarp: I think I agree, but I'm not sure. In a practical example of development, other packages are going to be added to this file over the lifetime of an app. Since there are multiple installs taking place, cleaning the apt-cache should happen below all of them.

I could add it to the bottom of the dockerfile, but I'm not sure it's worth it. At some point someone is going to need to shell into the container and check on something and if they need to install it'll be two steps rather than just one. Leaving it out means their debugging might be faster.

Since this is a development container, is the space savings worth it?

src/web_app_skeleton/docker/dev_entrypoint.sh
robacarp: Prebuilding the migrations in a release container is a great idea -- I do this in several apps and it's very handy.

src/web_app_skeleton/docker/dev_entrypoint.sh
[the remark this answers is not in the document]
robacarp: I mounted this stuff at boot time rather than at build time because I want to be able to capture the output of these commands in the docker volume so that a container rebuild doesn't require re-downloading the modules/shards. In a release image I would absolutely put them in the build phase.

src/web_app_skeleton/docker/dev_entrypoint.sh
robacarp: You're right that this error message is not right, and it's probably not necessary.

src/web_app_skeleton/docker-compose.yml
robacarp: Since compose is overlaying the app source code at runtime onto the built container image, these need to be volumes. If the system architecture (or operating system) differs from the container, node packages which have a compiled component will fail with weird and difficult to comprehend error messages when they're imported into the container.

src/web_app_skeleton/docker/development.dockerfile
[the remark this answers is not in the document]
grepsedawk: The apt update and apt install's definitely need to be combined, otherwise if you cache bust the install layer without busting the update layer, you're gonna have a bad time

PR #738, opened 2022-03-04
[Development/GettingStarted] Docker compose workflow

robacarp:
This is a series of contributions aimed at making getting started writing a Lucky app easier.

This allows an app generated with lucky_cli to play well in a `docker compose` based development environment. There are some benefits and drawbacks to this, but primarily it means that crystal, libssl, postgres, and other libraries/tools _don't need to be installed_ in order to get started. 

This does **not** provide a Dockerfile for production use, which would be very different from a development environment. After this is merged, I or someone else can tackle that, perhaps with a lot of inspiration from [Lucky Jumpstart](https://github.com/stephendolan/lucky_jumpstart/blob/main/dockerfiles/webserver.Dockerfile) or [my own lucky template](https://github.com/robacarp/lucky_template/blob/master/Dockerfile).

In order to get started with this, a generated app need only run `docker compose up` or `docker-compose up`. It will download images, create the database, install npm modules, install shard dependencies, and even migrate the database (just on the first boot).

### DIFF
--- a/src/web_app_skeleton/docker/dev_entrypoint.sh
+++ b/src/web_app_skeleton/docker/dev_entrypoint.sh
@@ -29,7 +29,7 @@ if ! [ -d bin ] ; then
 fi
 
 echo "Installing npm packages..."
-npm install
+yarn install
 
 if ! shards check ; then
   echo "Installing shards..."
--- a/src/web_app_skeleton/docker/dev_entrypoint.sh
+++ b/src/web_app_skeleton/docker/dev_entrypoint.sh
@@ -37,9 +37,9 @@ if ! shards check ; then
 fi
 
 echo "Waiting for postgres to be available..."
-./docker/wait-for-it.sh postgres:5432
+./docker/wait-for-it.sh -q postgres:5432
 
-if ! psql -d "$DATABASE_URL" -c '\d migrations' ; then
+if ! psql -d "$DATABASE_URL" -c '\d migrations' > /dev/null ; then
   echo "Finishing database setup..."
   lucky db.migrate
 fi
--- a/src/web_app_skeleton/docker/dev_entrypoint.sh
+++ b/src/web_app_skeleton/docker/dev_entrypoint.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# This is the entrypoint script used for development docker workflows. By
+# default it will:
+#  - Install shards.
+#  - Run migrations.
+#  - Start the dev server.
+# It also accepts any commands to be run instead.
+
+
+warnfail () {
+  echo "$@" >&2
+  exit 1
+}
+
+case ${1:-} in
+  "") # If no arguments are provided, start lucky dev server.
+    ;;
+
+  *) # If any arguments are provided, execute them instead.
+    exec "$@"
+esac
+
+echo "Installing npm packages..."
+npm install
+
+if ! shards check ; then
+  echo "Installing shards..."
+  shards install
+fi
+
+echo "Waiting for postgres to be available..."
+./docker/wait-for-it.sh postgres:5432
+
+echo "Starting lucky dev server..."
+exec lucky dev || warnfail "Lucky dev server failed."
--- a/src/web_app_skeleton/docker/dev_entrypoint.sh
+++ b/src/web_app_skeleton/docker/dev_entrypoint.sh
@@ -44,5 +44,10 @@ if ! psql -d "$DATABASE_URL" -c '\d migrations' > /dev/null ; then
   lucky db.migrate
 fi
 
+if [ -S .overmind.sock ] ; then
+  echo "Removing old overmind socket file..."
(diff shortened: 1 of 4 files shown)


src/web_app_skeleton/docker/development.dockerfile
```diff
@@ -0,0 +1,36 @@
+FROM crystallang/crystal:1.1.1
```
robacarp, line 1: This needs to be updated, but later versions of the crystal dockerfile don't play well with my computer yet. If someone is willing to test this on an AMD64 architecture with this line updated to 1.3.2, I'll be happy to update it.

src/web_app_skeleton/docker/wait-for-it.sh
```diff
@@ -0,0 +1,189 @@
+#!/usr/bin/bash
+#
+# Pulled from https://github.com/vishnubob/wait-for-it on 2022-02-28.
+# Licensed under the MIT license as of 81b1373f.
+#
+# Below this line, wait-for-it is the original work of the author.
+#
+# Use this script to test if a given TCP host/port are available
+
+WAITFORIT_cmdname=${0##*/}
```
jwoertink, line 10: waiting

src/web_app_skeleton/docker/development.dockerfile
```diff
@@ -0,0 +1,36 @@
...
+#  - Yarn is the default package manager for the node component of a lucky
+#    browser app.
+#  - Mix is the default asset compiler.
+RUN npm install -g yarn mix
+
+# Installs overmind, not needed if nox is the process manager.
+RUN wget https://github.com/DarthSim/overmind/releases/download/v2.2.2/overmind-v2.2.2-linux-amd64.gz && \
+    gunzip overmind-v2.2.2-linux-amd64.gz && \
+    mv overmind-v2.2.2-linux-amd64 /usr/bin/overmind && \
+    chmod +x /usr/bin/overmind
```
jwoertink, line 24: Should we be using [Nox](https://github.com/luckyframework/lucky_cli/blob/009f22ca4b9cdb12d807146b753a6860dc8fe91a/shard.yml#L22-L24) for this? It comes with Lucky CLI which gets installed in the next run.

src/web_app_skeleton/docker/dev_entrypoint.sh
```diff
@@ -0,0 +1,48 @@
...
+  "") # If no arguments are provided, start lucky dev server.
+    ;;
+
+  *) # If any arguments are provided, execute them instead.
+    exec "$@"
+esac
+
+if ! [ -d bin ] ; then
+  echo "Creating bin directory"
+  mkdir bin
```
jwoertink, line 28: does this need the `-p` flag?

src/web_app_skeleton/docker/dev_entrypoint.sh
```diff
@@ -0,0 +1,48 @@
...
+fi
+
+echo "Installing npm packages..."
+npm install
```
jwoertink, line 32: Why `npm` here and not `yarn`?

jwoertink (commented):
I love the direction! I just had a few questions, but no real changes that I can see. Thanks for putting this together :rocket:

src/web_app_skeleton/docker/development.dockerfile
robacarp: Yep. This provides overmind in the system, but if lucky-cli doesn't need it, it won't hurt anything. I needed this because I developed it against lucky-cli release rather than main.

src/web_app_skeleton/docker/dev_entrypoint.sh
robacarp: It shouldn't. It's only one level deep (rather than `mkdir one/two/three/bin`) and it'll only get run if the bin directory doesn't already exist.

I honestly don't think this stanza should need to be in here. The compiler should be creating this folder when the binary is created, but it doesn't. For reasons I don't understand the `ld` step  fails because the bin directory doesn't exist. This may be a linux problem I'm unaware of, but I was able to work around it with this.

src/web_app_skeleton/docker/dev_entrypoint.sh
robacarp: Oh good call, whooooooooopsiedaisy.

src/web_app_skeleton/docker/development.dockerfile
jwoertink: Ah. I forgot Nox integration hasn't been released yet. Yeah, the next release won't need a process runner!

robacarp: Since this will ship with Nox, when the cli is revised, I'm happy to pull it now or leave it in. I don't think it hurts and it might help someone get a compose going for an app that doesn't yet have the newest version of lucky on their system.

matthewmcgarvey (approved):

PR #739, opened 2022-03-05
Comment out unconfigured RequestIdHandler

grepsedawk:
A new lucky app results in ameba errors due to this block's contents
being commented out. By commenting out the whole block, we still
provide the user useful configuration information without leaving an
empty block


jwoertink (approved):
Sounds good :+1:  Thanks!

PR #742, opened 2022-03-19
Change default generated app to build a generic target

### LINKED ISSUE #735: Make shard.yml target generic
In shard.yml, the main target results in your app's constant name.
This means everybody's production config needs to call a different binary, which comes into play with Dockerfiles

I have a slight hack to get around this:
```bash
mv ./bin/`ls ./bin/ | head -1` /usr/local/bin/webserver
```
https://github.com/superfly/flyctl/pull/852/files#diff-338317355cac9afd0ed67e6eb8e770bd64ea5c80b801fa787af9c99511761d42R30

But it would be better if the target was just consistently named `webapp`

grepsedawk:
This allows for easier devops/DX by being able to have a reliable bin/app generated on build
Which greatly cleans up universal Dockerfiles

Resolves #735

### DIFF
--- a/spec/integration/init_web_spec.cr
+++ b/spec/integration/init_web_spec.cr
@@ -16,7 +16,6 @@ describe "Initializing a new web project" do
 
       File.delete("test-project/.env")
       compile_and_run_specs_on_test_project
-      File.read("test-project/Procfile").should contain "test_project"
       File.read(".github/workflows/ci.yml").should contain "postgres"
       File.read("test-project/public/mix-manifest.json").should contain "images/cat.gif"
       File.exists?("test-project/public/favicon.ico").should eq true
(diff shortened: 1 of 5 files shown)


jwoertink (commented):
I think there's a few other places this has to be changed. Like https://github.com/luckyframework/lucky_cli/blob/0b5b8fb25d8fc45d3cbc0334853daf45a1075d87/src/web_app_skeleton/Procfile.ecr#L1 for example

Also, I forget, but maybe the Heroku buildpack may also need to be updated? I'll try to dig around the ecosystem a bit and see where all this gets used. Thanks for getting this started!

grepsedawk:
I'll `grep` for project_name and see if I find other places.

grepsedawk:
Ooh nice, one .ecr taken away, certainly a win in my eyes
@jwoertink didn't find any other refs that need to change

grepsedawk:
Oop, forgot to mark this as ready for review

jwoertink (approved):
:+1:

PR #743, opened 2022-03-19
Generate apps with SecTester

### LINKED ISSUE #740: Adding option for integrated sec tester
@bararchy and I have been working on getting a smooth integration with the [SecTester](https://github.com/NeuraLegion/sec_tester) in to Lucky. I will be working on adding a config flag to generate an app with this setup out of the box. I just wanted to open an issue to track it.

jwoertink:
Fixes #740

Right now this is in WIP to get things rolling, but a few questions still need answering...

/cc. @bararchy

### DIFF
--- a/src/app_with_sec_tester/spec/flows/security_spec.cr.ecr
+++ b/src/app_with_sec_tester/spec/flows/security_spec.cr.ecr
@@ -1,0 +1,121 @@
+{% skip_file unless flag?(:with_sec_tests) %}
+# Run these specs with `crystal spec -Dwith_sec_tests`
+
+require "../spec_helper"
+
+describe "SecTester" do
+  <%- if browser? -%>
+  <%- if generate_auth? -%>
+  it "tests the sign_in for dom based XSS" do
+    with_cleanup(scanner) do
+      target = scanner.build_target(SignIns::New)
+      scanner.run_check(
+        scan_name: "ref: #{ENV["GITHUB_REF"]?} commit: #{ENV["GITHUB_SHA"]?} run id: #{ENV["GITHUB_RUN_ID"]?}",
+        tests: "dom_xss",
+        target: target
+      )
+    end
+  end
+
+  it "tests the sign_in page for SQLi, OSI, XSS attacks" do
+    with_cleanup(scanner) do
+      target = scanner.build_target(SignIns::Create) do |t|
+        t.body = "user%3Aemail=test%40test.com&user%3Apassword=1234"
+      end
+      scanner.run_check(
+        scan_name: "ref: #{ENV["GITHUB_REF"]?} commit: #{ENV["GITHUB_SHA"]?} run id: #{ENV["GITHUB_RUN_ID"]?}",
+        tests: [
+          "sqli",
+          "osi",
+          "xss",
+        ],
+        target: target
+      )
+    end
+  end
+
+  it "tests the sign_up page for dom based XSS" do
+    with_cleanup(scanner) do
+      target = scanner.build_target(SignUps::New)
+      scanner.run_check(
+        scan_name: "ref: #{ENV["GITHUB_REF"]?} commit: #{ENV["GITHUB_SHA"]?} run id: #{ENV["GITHUB_RUN_ID"]?}",
+        tests: "dom_xss",
+        target: target
+      )
+    end
+  end
+
+  it "tests the sign_up page for ALL attacks" do
+    with_cleanup(scanner) do
+      target = scanner.build_target(SignUps::Create) do |t|
+        t.body = "user%3Aemail=aa%40aa.com&user%3Apassword=123456789&user%3Apassword_confirmation=123456789"
+      end
+      scanner.run_check(
+        scan_name: "ref: #{ENV["GITHUB_REF"]?} commit: #{ENV["GITHUB_SHA"]?} run id: #{ENV["GITHUB_RUN_ID"]?}",
+        tests: nil,
+        target: target
+      )
+    end
+  end
+  <%- end -%>
+  it "tests root for header security issues" do
+    with_cleanup(scanner) do
+      target = scanner.build_target
+      scanner.run_check(
+        scan_name: "ref: #{ENV["GITHUB_REF"]?} commit: #{ENV["GITHUB_SHA"]?} run id: #{ENV["GITHUB_RUN_ID"]?}",
+        tests: "header_security",
+        target: target
+      )
+    end
+  end
+
+  it "tests root for cookie security issues" do
+    with_cleanup(scanner) do
+      target = scanner.build_target
+      scanner.run_check(
+        scan_name: "ref: #{ENV["GITHUB_REF"]?} commit: #{ENV["GITHUB_SHA"]?} run id: #{ENV["GITHUB_RUN_ID"]?}",
+        tests: "cookie_security",
+        target: target
+      )
+    end
+  end
+
+  it "tests app.js for 3rd party issues" do
+    with_cleanup(scanner) do
+      target = SecTester::Target.new(Lucky::AssetHelpers.asset("js/app.js"))
+      scanner.run_check(
+        scan_name: "ref: #{ENV["GITHUB_REF"]?} commit: #{ENV["GITHUB_SHA"]?} run id: #{ENV["GITHUB_RUN_ID"]?}",
+        tests: "retire_js",
+        target: target
+      )
+    end
+  end
+  <%- else -%>
+  it "tests the sign_in API for SQLi, OSI, XSS attacks" do
+    with_cleanup(scanner) do
+      target = scanner.build_target(Api::SignIns::Create) do |t|
+        t.body = "user%3Aemail=test%40test.com&user%3Apassword=1234"
+      end
+      scanner.run_check(
+        scan_name: "ref: #{ENV["GITHUB_REF"]?} commit: #{ENV["GITHUB_SHA"]?} run id: #{ENV["GITHUB_RUN_ID"]?}",
+        tests: [
+          "sqli",
+          "osi",
+          "xss",
+        ],
+        target: target
+      )
+    end
+  end
+  <%- end -%>
+end
+
+private def scanner
+  LuckySecTester.new
+end
+
+private def with_cleanup(tester : LuckySecTester)
+  yield
+ensure
+  tester.try &.cleanup
+end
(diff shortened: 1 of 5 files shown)


src/app_with_sec_tester/spec/flows/security_spec.cr.ecr
```diff
@@ -0,0 +1,121 @@
...
+        scan_name: "ref: #{ENV["GITHUB_REF"]?} commit: #{ENV["GITHUB_SHA"]?} run id: #{ENV["GITHUB_RUN_ID"]?}",
+        tests: nil,
+        target: target
+      )
+    end
+  end
+  <%- end -%>
+  it "tests root for header security issues" do
+    with_cleanup(scanner) do
+      target = scanner.build_target
```
jwoertink, line 63: @bararchy when you wrote "root" here, is this testing the home page? I'm wondering if we should specify the Home action, or does this just test the whole app? Same with the cookie one below....

src/app_with_sec_tester/spec/flows/security_spec.cr.ecr
```diff
@@ -0,0 +1,121 @@
...
+        target: target
+      )
+    end
+  end
+  <%- else -%>
+  it "tests the sign_in API for SQLi, OSI, XSS attacks" do
+    with_cleanup(scanner) do
+      target = scanner.build_target(Api::SignIns::Create) do |t|
+        t.body = "user%3Aemail=test%40test.com&user%3Apassword=1234"
+      end
```
jwoertink, line 98: We never tested API only in https://github.com/bararchy/lucky_sec_test. This body would need to be JSON, and then https://github.com/luckyframework/lucky_sec_tester/blob/205dcc0e7e1ebfa9548219077c95121e8a5bf795/src/lucky_sec_tester.cr#L42 would need to be updated to know we're using JSON here...

src/app_with_sec_tester/spec/setup/sec_tester.cr
```diff
@@ -0,0 +1,8 @@
+require "lucky_sec_tester"
+
+# Signup for a `NEXTPLOT_TOKEN` at
+# [NeuraLegion](https://app.neuralegion.com/signup)
```
jwoertink, line 4: The company name is now called [Bright](https://brightsec.com/), but as of this comment, the Github org and this URL are still pointed to NeuraLegion... I don't want to confuse people, so this will need to change once the company's name is fully moved over.

src/generators/web.cr
```diff
@@ -189,6 +199,14 @@ class LuckyCli::Generators::Web
           version: ~> 0.7.3
       DEPS_LIST
     end
+
+    if with_sec_tester?
+      append_text to: "shard.yml", text: <<-DEPS_LIST
+        lucky_sec_tester:
```
jwoertink, line 205: oops... just realized if you do api_only and with_sec_tester, it'll append this to dependencies and not development dependencies... My comment [here](https://github.com/luckyframework/lucky_cli/issues/735#issuecomment-1073047223) on just requiring yaml is starting to sound a lot nicer....

grepsedawk: Alternatively, we could forcibly add a devdep section and use "append before" and "append after" to add dev deps vs dev deps

src/app_with_sec_tester/spec/flows/security_spec.cr.ecr
bararchy: Oh, so root as in like "root directory" so yeha it's the home page. if there is a HOME action that could be cool to add :)

src/app_with_sec_tester/spec/flows/security_spec.cr.ecr
bararchy: yeha, we should set a default but allow users to overload?

bararchy:
@jwoertink if we need to configure the API key for Lucky let me know and I can make sure the Lucky Org in Bright app has multiple engines to run all needed parallel scans :)

grepsedawk:
Is SecTester free for private source?

bararchy:
@grepsedawk yeha :) you can signup for an API key for free https://app.neuralegion.com/signup

src/web_app_skeleton/.github/workflows/ci.yml.ecr
```diff
@@ -117,4 +117,4 @@ jobs:
           ./lucky_tasks db.seed.required_data
 
       - name: Run tests
-        run: crystal spec
+        run: crystal spec <%- if with_sec_tester? %>-Dwith_sec_tests<%- end -%>
```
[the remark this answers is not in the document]
bararchy: @jwoertink yeha, once it's in I'm sure we can get it working :) 
Let me know if I can help

spec/integration/init_web_spec.cr
```diff
@@ -17,7 +17,7 @@ describe "Initializing a new web project" do
       File.delete("test-project/.env")
       compile_and_run_specs_on_test_project
       File.read("test-project/Procfile").should contain "test_project"
-      File.read(".github/workflows/ci.yml").should contain "postgres"
+      File.read("test-project/.github/workflows/ci.yml").should contain "postgres"
```
jwoertink, line 20: turns out this has been wrong for a while :sweat_smile:

bararchy:
@jwoertink I saw you're using node12 which should be OK, I verified in the docs that it's supported: https://docs.brightsec.com/docs/installation-options#npm 

So anything 10-14 should be good, I also looked for the `touch` thingy but I can't find it. One reason for process to not be up is something in the ENV? I'll try and add more logs. can you check locally with the API key and with `LOG_LEVEL=DEUBG`? see that at least locally we get it working

jwoertink:
That touch happens on the LuckySecTester shard itself Since that shard doesn't actually need to run any nexploit checks.

https://github.com/luckyframework/lucky_sec_tester/blob/afc4d99ee1b8fbd371a2b9b0a5f0ac1bee185859/.github/workflows/ci.yml#L49-L50

I can use a newer version of node, but it just seems like the `scanner.client` isn't being called or something. I'll try locally and see what I can find out.

bararchy:
@jwoertink can we try and re-run using latest https://github.com/luckyframework/lucky_sec_tester/commit/346ceb3567110e433863dd1997afcec6243db3e5 ?

src/app_with_sec_tester/spec/flows/security_spec.cr.ecr
```diff
@@ -0,0 +1,117 @@
...
+      )
+    end
+  end
+  <%- else -%>
+  <%- if generate_auth? -%>
+  it "tests the sign_in API for SQLi, and JWT attacks" do
+    scan_with_cleanup do |scanner|
+      api_headers = HTTP::Headers{"Content-Type" => "application/json", "Accept" => "application/json"}
+      target = scanner.build_target(Api::SignIns::Create, headers: api_headers) do |t|
+        t.body = {"user" => {"email" => "aa%40aa.com", "password" => "123456789"}}.to_json
```
bararchy, line 92: @jwoertink as this is a JSON API call, do we actually need to encode the `@` here?

jwoertink: probably not....

jwoertink:
it's green!!!

.github/workflows/ci.yml
```diff
@@ -65,9 +65,6 @@ jobs:
         uses: crystal-lang/install-crystal@v1
         with:
           crystal: ${{ matrix.crystal_version }}
-      - uses: actions/setup-node@v3.1.0
-        with:
-          node-version: "12.x"
```
jwoertink, line 70: Just use the node that comes installed on ubuntu-latest

.github/workflows/weekly.yml
```diff
@@ -1,9 +1,6 @@
 name: Lucky CLI Weekly CI
 
 on:
-  push:
-    branches: [main]
-  pull_request:
```
jwoertink, line 6: This was causing this to run on every single commit. It should be a separate PR, but I needed it to quiet down with all the testing.

.github/workflows/weekly.yml
```diff
@@ -44,6 +41,9 @@ jobs:
       - uses: actions/setup-node@v3.1.0
         with:
           node-version: "12.x"
```
jwoertink, line 43: this actually needs to go away

spec/integration/sec_tester_spec.cr
```diff
@@ -0,0 +1,23 @@
+require "../spec_helper"
+
+{% if env("RUN_SEC_TESTER_SPECS") == "1" %}
```
jwoertink, line 3: Anyone that wants to run this spec locally would need a Bright API key... That's not really going to happen, so we can treat this like the Heroku specs. As long as this spec passes in CI, and the rest pass locally, we should be fine

bararchy (approved):

bararchy:
Amazing work @jwoertink! :green_circle: light!

src/web_app_skeleton/script/setup.ecr
```diff
@@ -22,7 +22,7 @@ print_done
 <%- end -%>
 
 notice "Installing shards"
-shards install | indent
+shards install --ignore-crystal-version | indent
```
robacarp, line 25: Is this a "hey i need this to work for now" contribution which should be removed? Or should this be kept in?

robacarp (approved):
This seems fine. The way this pattern scales makes me nervous, but I get that we need to have a few dice in the bucket before we can have enough to roll yahtzee.

src/web_app_skeleton/script/setup.ecr
jwoertink: This was a tough one... Without it, all tests again < 1.2.2 fail because of https://github.com/luckyframework/lucky_sec_tester/blob/346ceb3567110e433863dd1997afcec6243db3e5/shard.yml#L7 That's an easy one to drop down, but then we can only drop down to 1.1.1 https://github.com/NeuraLegion/sec_tester/blob/9db160150b4cc15ebb6d6722a25ee55a5c5af3ce/shard.yml#L7 and CLI tests against 1.0.0 https://github.com/luckyframework/lucky_cli/blob/89be92657d24128cd2623198e62c13bed1a3ee1b/.github/workflows/ci.yml#L35

Do I just raise the min test to 1.1.1 for CLI since 1.4 comes out tomorrow? Then I could update the other shard, and remove this

jwoertink: Alright, I updated the LuckySecTester to be >=1.1.1. It would still fail due to checking 1.0.0. So I still have to either use the flag, or drop 1.0.0 support :thinking:  I'm sort of ok with going with the latter. I feel like once 1.4.0 drops, few people will be wanting to upgrade Lucky and continue using 1.0.0....

robacarp: I don't have any reason _not_ to have this flag in there, I was just curious. It's complicated, and I get that.

PR #755, opened 2022-04-26
Fix and Optimize tests (2)

bararchy:
Trying to see if we can make CI run


bararchy:
@jwoertink we are golden it seems :grin:

jwoertink (approved):
sweet! Magic combo :boom:

PR #756, opened 2022-04-27
Update ci.yml

bararchy:
remove duplicate run


jwoertink (approved):
So strange that one allows secret keys to be used, and one doesn't...

PR #775, opened 2022-09-09
update the Dockerfile to include utilites needed to make it run

jwoertink:
Fixes https://github.com/luckyframework/lucky/issues/1732

### DIFF
--- a/src/web_app_skeleton/docker/development.dockerfile
+++ b/src/web_app_skeleton/docker/development.dockerfile
@@ -1,5 +1,9 @@
 FROM crystallang/crystal:1.4.1
 
+# Install utilities required to make this Dockerfile run
+RUN apt-get update && \
+    apt-get install -y wget gunzip
+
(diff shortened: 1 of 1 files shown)


src/web_app_skeleton/docker/development.dockerfile
```diff
@@ -1,5 +1,9 @@
 FROM crystallang/crystal:1.4.1
 
+# Install utilities required to make this Dockerfile run
+RUN apt-get update && \
+    apt-get install -y wget gunzip
```
matthewmcgarvey, line 5: Can this be moved down to the other section using `apt-get`?

jwoertink: The line right after this runs `wget`, so it has to be installed before we get to that. Then that one is required before the following apt-get so node can be installed :sweat_smile:

matthewmcgarvey (approved):

PR #776, opened 2022-09-10
Problem when executing lucky db.migrate

wilful:
I'm not a Crystal specialist, but there seems to be a typo here

```bash
web          | Showing last frame. Use --error-trace for full trace.
web          |
web          | In src/shards.cr:7:1
web          |
web          |  7 | require "avram/lucky"
web          |      ^
web          | Error: can't find file 'avram/lucky'
web          |
web          | If you're trying to require a shard:
web          | - Did you remember to run `shards install`?
web          | - Did you make sure you're running the compiler in the same directory as your shard.yml?
web          | Done
```

### DIFF
--- a/src/web_app_skeleton/src/shards.cr.ecr
+++ b/src/web_app_skeleton/src/shards.cr.ecr
@@ -4,7 +4,7 @@ LuckyEnv.load?(".env")
 
 # Require your shards here
 require "lucky"
-require "avram/lucky"
+require "avram"
 require "carbon"
 <%- if generate_auth? -%>
 require "authentic"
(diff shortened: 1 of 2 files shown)


jwoertink:
This isn't a typo, this change will be required for the next Lucky release. The current main branch of the CLI isn't in a usable state yet. If you need to build the latest version of Lucky, you'll need to checkout the 0.30 tag, and build from that.

I'll try to get this cleaned up today if I have some time.

wilful:
I'm used example from docs https://luckyframework.org/guides/getting-started/installing#install-lucky-cli-manually
```
git checkout v0.30.0
```

Issue #629, opened 2021-04-16
Consider replacing turbolinks with hotwire


matthewmcgarvey:
I'm not sure if we want to add all of hotwire or just turbo since that has turbolinks in it.

stephendolan:
I'm going to give https://github.com/jgaskins/turbo a try in my latest app, which will tangentially give me some experience with the JS portion as well.

One important thing to note is that we'll lose our `disable_with` helpers if we make this move.

snadon:
I'm replacing `turbolinks` with `turbo` right now. AFAIK, `rails_ujs` still works as long as you remove all your `data-remote`.

You'll need to add `data_turbo: false` to your forms otherwise you'll get an error because turbo expects a 400 or a redirect and not a 200. Currently we cannot use another status code when rendering a page (https://github.com/luckyframework/lucky/issues/1312).

rmarronnier:
I'm bumping this since luckyframework/lucky#1312 is fixed and hotwire/turbo has matured a lot since.

PR #779, opened 2022-09-27
Ripping out Turbolinks now that it's been deprecated and archived.

jwoertink:
Ref #629

We've discussed replacing turbolinks with turbo, but for now I'm just removing turbolinks as default now that the [repo](https://github.com/turbolinks/turbolinks-classic) has been marked as archived. Existing projects using it will still have the ability to use it, but new generated projects won't come with it by default. That will allow us to figure out how to add in turbo (if that's what we want....). Though, Lucky is pretty fast without it :joy:


jwoertink:
Merging this as I have another PR to pull everything together.

PR #780, opened 2022-09-28
Release 1.0.0-rc1

jwoertink:


jwoertink:
The edge specs won't pass unless I update the Wordmith version on main, but those updates are in the release update. Managing versions is hard :roll_eyes:

jwoertink:
woot! All green!

PR #800, opened 2023-05-14
Remove the crystal lib generator code.

### LINKED ISSUE #741: Consider dropping crystal init during setup
Right now when you run `lucky init`, we run `crystal init app NAME` under the hood. It's nice because it gives us a default structure, and a ton of things like setting up the src, and spec folders and such.

The issue is right after we do that, we basically remove almost all of it.

https://github.com/luckyframework/lucky_cli/blob/0b5b8fb25d8fc45d3cbc0334853daf45a1075d87/src/generators/web.cr#L107-L138

We used to even have this https://github.com/luckyframework/lucky_cli/commit/45f35290dae282610ea6aa5012b31861c792a03e because Crystal would default generate a travis CI, but then we started using Github, and then later versions of Crystal stopped generating that.

We're also at the mercy of whatever naming conventions, or bugs may appear in the crystal init.( [ref](https://github.com/crystal-lang/crystal/pull/11170) )

Related: https://github.com/luckyframework/lucky_cli/issues/735

In regards to that issue, if we just generated our own structure out of the box, we'd keep control and flexibility to what goes in to an app. 

The main downside would be that this also opens us to bugs with all the different permutations of app type (i.e. browser, api, auth, etc...), it's possible we miss something, or similar to the Crystal naming bug, we end up creating our own bugs with naming....

jwoertink:
Fixes #741
Ref https://github.com/luckyframework/lucky/discussions/1812

This PR removes the initial step of running `crystal init app ...`. We had been using this command to bootstrap the structure, but then remove almost every file after it was generated.

This now generates an empty directory, and only inserts what is needed which, in theory, should reduce file operations, as well as help in the effort to running on Windows.

/cc @mdwagner

### DIFF
--- a/src/generators/web.cr
+++ b/src/generators/web.cr
@@ -1,5 +1,3 @@
-require "option_parser"
-
 class LuckyCli::Generators::Web
   include LuckyCli::GeneratorHelpers
 
@@ -45,13 +43,9 @@ class LuckyCli::Generators::Web
 
   def run : Nil
     ensure_directory_does_not_exist
-    generate_default_crystal_project
-    rename_shard_target_to_app
-    add_deps_to_shard_file
-    add_dev_deps_to_shard_file
-    remove_generated_src_files
-    remove_generated_spec_files
-    remove_default_readme
+    generate_default_project
+    generate_shard_yml
+
     add_default_lucky_structure_to_src
 
     if browser?
@@ -72,7 +66,6 @@ class LuckyCli::Generators::Web
     end
 
     setup_gitignore
-    remove_default_license
 
     puts <<-TEXT
     #{"Done generating your Lucky project".colorize.bold}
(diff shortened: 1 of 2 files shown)


src/generators/web.cr
```diff
@@ -89,171 +82,80 @@ class LuckyCli::Generators::Web
     if default_directory?
       project_dir
     else
-      [full_project_directory.chomp('/'), project_dir].join('/')
+      File.join(full_project_directory.chomp('/'), project_dir)
```
jwoertink, line 85: I'm not sure how this looks on Windows, by we may need to see if there's some sort of `Crystal::System::FileSeparator` or whatever. I'm assuming on Windows we'd need to `chomp('\')`...

jwoertink:
Looks like the ameba failure is actually caused from https://github.com/crystal-ameba/ameba/issues/372 so it'll be fixed on the next release.

Not sure why the sectester stuff is failing... Seems it can't connect? @bararchy do you have any ideas on https://github.com/luckyframework/lucky_cli/actions/runs/4973638473/jobs/8899605681 ?

matthewmcgarvey (approved):

bararchy:
@jwoertink seems strange... do we maybe need to install `openssl` or `ca-certificate` ubuntu package? :shrug: 
Does it pass locally for you? because for me it works fine...

src/generators/web.cr
mdwagner: While you _can_ use `File` methods for dealing with `Path`s, because they are just wrappers around `Path`, I'd recommend to just use `Path` directly instead. It makes it more clear what you're doing (e.g. path manipulation), and use `File` for actual filesystem operations (e.g. read/write a file).

You don't actually need `#chomp` here:
```suggestion
      File.join(full_project_directory, project_dir)
```

But this is my suggestion (which is the same as `File#join`, just prefer using `Path`):
```suggestion
      Path.new(full_project_directory).join(project_dir).to_s
```

However, my above suggestion would be better suited in a separate PR to change _all other_ places using `File` instead of `Path` (I might add this to my lucky_template work).

jwoertink: Yeah, I never thought about using `Path` until I started seeing your code. I think that makes more sense. Maybe I'll just merge this PR in as-is, then in another one we can go through it all and really start cleaning stuff up.

PR #802, opened 2023-05-14
If the CSRF is disabled, also remove the metatags.

### LINKED ISSUE #696: CSRF should be removed if it's disabled in forms
You have the option to disable CSRF in forms with
https://github.com/luckyframework/lucky/blob/60b24a8f0489d6115063d15ba5974481b02c9abe/src/lucky/tags/form_helpers.cr#L3

If this is disabled, then we should probably also display the metatags as well. 

https://github.com/luckyframework/lucky_cli/blob/522c046f6cc3306c416a5c5175d589d5fd8bc682/src/browser_app_skeleton/src/components/shared/layout_head.cr#L11

jwoertink:
Fixes #696

This is to coincide with 
https://github.com/luckyframework/lucky/blob/22fcb187c2045a36823a03310d1ed68aec5d8f0b/src/lucky/tags/form_helpers.cr#L1C1-L8

### DIFF
--- a/src/browser_app_skeleton/src/components/shared/layout_head.cr
+++ b/src/browser_app_skeleton/src/components/shared/layout_head.cr
@@ -7,12 +7,27 @@ class Shared::LayoutHead < BaseComponent
       title "My App - #{@page_title}"
       css_link asset("css/app.css")
       js_link asset("js/app.js"), defer: "true"
-      csrf_meta_tags
+      csrf_meta_tags if include_csrf_tag?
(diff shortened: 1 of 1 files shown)


jwoertink:
Just realized that there's this too https://github.com/luckyframework/lucky/pull/1657

So what happens if you disable this, and then submit the form? My guess is it would fail because `allow_forgery_protection` would still be true :thinking:  Would we have to disable that too?

PR #804, opened 2023-06-10
Migrate to LuckyTemplate - Part 1 of ...

mdwagner:
- Changes app_with_sec_tester to use lucky_template
- Changes api_authentication_app_skeleton to use lucky_template
- ~Updates CI workflow to specifically run each expected matrix build (at least for now, it makes it more clear what things we expect to run. eventually, I hope we can just use matrix vars normally)~
- Didn't add specs for switching to lucky_template, as I plan to add them after subsequent PR. I expect there might be a small refactor
- Updated Weekly CI builds to include edge + nightly, and removed those as requirements on PRs


jwoertink:
Sorry, I just merged https://github.com/luckyframework/lucky_cli/pull/800 :sweat_smile:  so it may be a bit wonky.

mdwagner:
@jwoertink are these failures warranted? Also, for the nightly build, what if we had that run on a cron for main (maybe once per week)?

jwoertink:
The Ameba one will hopefully be fixed when they do a new release. I'm not sure why SecTester is failing. I think we just need to release a new version of SecTester and make sure that's updated, but I wouldn't worry about it on this PR.

Also, for the nightly, moving to a weekly cron sounds good. It would be nice to have them checked on each PR, but since the Crystal nightlies don't actually build new releases nightly, I guess it doesn't make a ton of sense how we have it setup now.

jwoertink (approved):
Nice! This CLI will look completely different when these are all done :partying_face:

mdwagner:
@jwoertink 
So, I realized that there already was a Weekly cron CI, and just updated that.

Also, I think having edge and nightly builds on main actually makes more sense, since they are more likely to break, and I would think we don't want contributions to have such a high bar, unless the expectation is something like "this project only exists off nightly/latest", right?

I think PRs should be pretty consistent when run, and instead let advanced contributors deal with upcoming issues as they see fit.

PR #808, opened 2023-06-23
Migrate to LuckyTemplate - Part 2 of ...

mdwagner:
- Changes base_authentication_app_skeleton to use lucky_template
- Fixes regression of `src/generators/web.cr` from #800


jwoertink (approved):
It seems ok from what I can tell. We'll probably have to just get all of in and then test like crazy :joy:

PR #809, opened 2023-06-24
Migrate to LuckyTemplate - Part 3 of ...

mdwagner:
I decided to take a second look at `ShardFileGenerator`, and found that not only could I use `lucky_template`, it could also be cleaned up to leverage `Hash#to_yaml`.

### DIFF
--- a/src/lucky_cli/shard_file_generator.cr
+++ b/src/lucky_cli/shard_file_generator.cr
@@ -1,4 +1,5 @@
 require "yaml"
+require "lucky_template"
 
 class ShardFileGenerator
   getter project_name : String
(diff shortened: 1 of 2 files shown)


mdwagner:
Not sure why Security specs are failing...

src/lucky_cli/shard_file_generator.cr
```diff
@@ -14,91 +15,113 @@ class ShardFileGenerator
...
+  def shard_info : String
+    normalize_yaml({
+      "name"    => project_name,
+      "version" => "0.1.0",
+      "targets" => {
+        project_name => {
+          "main" => Path.new("src", "#{project_name}.cr").to_s,
+        },
+      },
+      "crystal" => ">= #{Crystal::VERSION}",
```
jwoertink, line 41: This is definitely easier to read than what I had :raised_hands:

jwoertink (approved):
Sweet :+1:  I wouldn't worry about the sectester for the moment. I think we have to do a new release, then update the CLI to pull that, but we can do that later.

Issue #778, opened 2022-09-17
Remove .keep files with other files in the directory during init process


matthewmcgarvey:
The most annoying thing that I find myself wasting time on when I create a new project is deleting these pesky `.keep` files from directories. They're all over and almost none of them are necessary in a normally generated project. It shouldn't be too hard to find all `.keep` files and delete any with other files in the directory. It could be one of the last steps in the init command and it would save me like 5 minutes of time 🙏

jwoertink:
Aren't they needed because default git setups won't include empty directories? So you end up losing context. Or is that not a thing anymore? 

Related: https://github.com/luckyframework/lucky_cli/issues/741

stephendolan:
Yeah, I the idea is that you preserve the expected structure of a Lucky app using these .keep files, even if you remove all of the non-hidden files (like clearing our all actions).

I don't think we should do this by default, but like the idea of having it as an init flag like Rails does (I think it's --no-keep).

matthewmcgarvey:
Sorry, it was late and I didn't explain it well. I'm talking about `.keep` files that are in directories with other files or directories. Because of that, they are there for no reason. So what i'm recommending is leaving the `.keep` files that are there for a reason and remove the ones that aren't (which is the vast majority on a normal app).

stephendolan:
Is the reason for those that we might add a generator to the CLI that doesn't populate them, though?

jwoertink:
Yeah, I guess that makes sense. If we generate a .keep inside of the src/actions, that should go away since you'll never have an empty actions directory in Lucky. It almost seems like it's just mixins directories that would have it, right?

PR #810, opened 2023-06-25
Migrate to LuckyTemplate - Part 4 of ...

mdwagner:
- Changes browser_authentication_app_skeleton to use lucky_template


src/lucky_cli/browser_authentication_src_template.cr
```diff
@@ -1,3 +1,167 @@
...
+
+  private def actions_folder
+    LuckyTemplate.create_folder do |actions_dir|
+      actions_dir.add_folder("me") do |me_dir|
+        me_dir.add_file("show.cr") do |io|
+          ECR.embed("#{__DIR__}/../browser_authentication_app_skeleton/src/actions/me/show.cr.ecr", io)
+        end
+      end
+      actions_dir.add_folder("mixins") do |mixins_dir|
+        mixins_dir.add_file(".keep")
```
jwoertink, line 45: Maybe we can knock out https://github.com/luckyframework/lucky_cli/issues/778 too while you're in here? This issue is basically saying no `.keep` inside folders that will always have files generated. We can also come back to that later once everything is in place.

src/lucky_cli/browser_authentication_src_template.cr
```diff
@@ -1,3 +1,167 @@
...
+        end
+        sign_ups_dir.add_file("new.cr") do |io|
+          ECR.embed("#{__DIR__}/../browser_authentication_app_skeleton/src/actions/sign_ups/new.cr.ecr", io)
+        end
+      end
+    end
+  end
+
+  private def emails_folder
+    LuckyTemplate.create_folder do |emails_dir|
```
jwoertink, line 117: That's interesting that this already knows what directory its in. Pretty neat!

jwoertink (approved):

src/lucky_cli/browser_authentication_src_template.cr
mdwagner: The only issue I have removing `.keep` files are that it removes the _intention_ of whether a particular folder structure should _always exist_, e.g. `mixins` is a folder name you _expect_ to see in a project. Or `support` is a folder name you _expect_ to see in `spec`. That's my opinion, otherwise I agree you don't need to include them (although, they don't cause any harm in having them).

src/lucky_cli/browser_authentication_src_template.cr
mdwagner: It's not that it already knows, it's more that I'm making it clear that I expect it to be this way. You still need to add it to `Folder#insert_folder` and provide a name (which could be something other than "emails"). This is just so you can separate out the folders for better code organization or for reuse.

src/lucky_cli/browser_authentication_src_template.cr
mdwagner: We could always add an option to the CLI that doesn't add those files, which seems like the better option here.

jwoertink: I'm cool with that too. The files being there don't really bother me. We can just leave it for now and worry about that in a future PR.

Issue #782, opened 2022-09-30
Embedded docs link needs url


robacarp:
https://github.com/luckyframework/lucky_cli/blob/d947fe68a0b29b16137101f6d5424af5933aba38/src/browser_app_skeleton/src/components/shared/layout_head.cr#L13-L16

jwoertink:
Related: https://github.com/luckyframework/website/issues/1071

robacarp:
On second thought I'm not convinced that the docblock even needs to be there. The rest of the helper methods in that component aren't supported by documentation. Perhaps the "only in development" should be made more explicit by doing something like:
 
```crystal
live_reload_connect_tag if Lucky::Env.development?
```

Or perhaps, alternatively, all of the method calls in that component should have a docblock that points to documentation?

PR #811, opened 2023-07-12
Remove live reload if

snadon:


src/browser_app_skeleton/src/components/shared/layout_head.cr
```diff
@@ -12,7 +12,7 @@ class Shared::LayoutHead < BaseComponent
 
       # Development helper used with the `lucky watch` command.
       # Reloads the browser when files are updated.
-      live_reload_connect_tag if LuckyEnv.development?
+      live_reload_connect_tag
```
jwoertink, line 15: Oh, we already had that? I totally forgot. I should have looked here first. So this goes back to https://github.com/luckyframework/lucky_cli/issues/782 I vaguely remember there being a few that was confused about this. Though the conditional is actually irrelevant, it seems it provides sort of "documentation" beyond the comment itself.

Here's the PR that added it in https://github.com/luckyframework/lucky_cli/pull/794

I'm not sure which is a better default :man_shrugging:

snadon: The fact that the Luckyframework website is different confused me.

Maybe we should add a comment on the same line?
```crystal
live_reload_connect_tag # automatically omitted for production
```

jwoertink: The website is probably different because that change was missed after the last release. I guess a comment could work too

snadon:
Closing since the decision was already made in the past and we have better things to do :)

PR #812, opened 2023-07-13
Migrate to Lucky Template - Part 5 of ...

mdwagner:


jwoertink (approved):
I pulled this down and was able to generate all of the different apps, boot them, run specs, and everything. I'm not sure what's left here, but it looks good to me! :+1:

mdwagner:
@jwoertink I remembered I wanted to remove teeplate since we don't need it anymore. So, this PR should be mergeable now.
(optional) I wanted to rewrite the tests using LuckyTemplate, but it's not required for a new release if you wanted to go ahead without it.

PR #815, opened 2023-09-21
Specs Overhaul + build system via Earthly

mdwagner:
**Highlights**
- Introduced _snapshots_ for testing if generated files and folders match their expectations (this includes content)
- Now using [Earthly](https://earthly.dev/) for both CI and local development
- Updated README with more current information
- Removed `script` folder, as Earthly provides a nicer and more complete interface

**Specs Overhaul**
- Testing is now split up into 3 parts: Unit, Integration, and End-to-end
- Unit
  - Requirements: `crystal`
  - Tests that run `crystal spec --tag "~integration"`
  - All supported platforms should be able to run these
- Integration
  - Requirements: `crystal`, `lucky`
  - Tests that run `crystal spec --tag integration` and need `lucky` in `PATH`
  - Only Linux runs these
- End-to-end
  - Requirements: `earthly`, `docker`
  - Test suites that run inside docker containers using Earthly
  - Only Linux runs these
- Security (also End-to-end)
  - Requirements: Bright secrets, `earthly`, `docker`
  - Similar to End-to-end test suite, but runs with BrightSec
  - Only Linux runs these


mdwagner:
@jwoertink This is an example of how I want to test out the generators. Let me know if you want me to clarify anything. Still WIP though.

mdwagner:
TODO: look at merging `lucky_spec` and `task_spec`

mdwagner:
TODO: write guide for snapshots, earthly, testing, etc.

jwoertink (approved):
It looks good from what I can tell. It might be a bit hard to really tell until we can get it in there and start using it.

One question I have is, with these files in the `fixtures`, is this where the templates all pull from, or are these duplicates of what goes in the templates just used for testing?

mdwagner:
They're duplicates in order to consistently know when things change. Which means, they are essentially the final versions of the files, but in a fossilized form. Also, they needed to be removed from the spec folder, because Crystal would try to include them, but they aren't meant to be used, just compared with.

mdwagner:
Consequently, because they are the final form, it's easier to see what it looks like from a user perspective.

jwoertink:
Sounds good. I saw we get this in and really start playing around with it to get a better feel for the flow.

PR #819, opened 2023-10-26
Pass custom task args to the tasks

### LINKED ISSUE #816: Passing help flag to any task now only shows lucky help menu
In https://github.com/luckyframework/lucky_cli/pull/771 I moved some code around so when you call `lucky -h` it doesn't start compiling. The side effect that I missed is now if you run `lucky exec -h` you get the `lucky -h` menu instead. 

This has to be fixed or reverted before we can do a release.

jwoertink:
Fixes #816 

When I added in https://github.com/luckyframework/lucky_cli/pull/771 It allowed you to run `lucky -h` without waiting on your app to compile. The side affect to this was when you try to pass args to a custom task, the `OptionParser` in here would swallow those and fail. Something as simple as `lucky db.create -h` would return the same help as `lucky -h`. 

In order to tell the OptionParser to stop parsing args, we can use the `parser.stop`. This essentially places a `--` in the spot where we stop the parser.

### DIFF
--- a/spec/integration/task_spec.cr
+++ b/spec/integration/task_spec.cr
@@ -32,7 +32,7 @@ private def run(process, output = STDOUT)
 end
 
 private def task(task_name)
-  run("crystal spec/tasks.cr --no-debug -- #{task_name}")
+  run("crystal run --no-debug spec/tasks.cr -- #{task_name}")
 end
 
 private def be_successful
--- a/src/lucky.cr
+++ b/src/lucky.cr
@@ -27,6 +27,10 @@ private def precompiled_task_path : String?
   end
 end
 
+private def built_in_commands : Array(String)
+  ["dev", "tasks", "init", "init.custom"]
+end
+
 private def print_lucky_app_not_detected_error
   puts <<-ERROR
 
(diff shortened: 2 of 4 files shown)


spec/integration/task_spec.cr
```diff
@@ -32,7 +32,7 @@ private def run(process, output = STDOUT)
 end
 
 private def task(task_name)
-  run("crystal spec/tasks.cr --no-debug -- #{task_name}")
+  run("crystal run --no-debug spec/tasks.cr -- #{task_name}")
```
jwoertink, line 35: We've had this "Task --no-debug not found" showing up for a while, and I just found it here.

src/lucky.cr
```diff
@@ -80,22 +84,31 @@ options = OptionParser.new do |parser|
...
-      }
-    }
+      end
+    end
+
+    if ARGV.size > 1 && !built_in_commands.includes?(task_name)
+      # This stops the CLI parsing options beyond this point.
+      # It allows args to be passed to custom CLI tasks
+      parser.stop
+    end
```
jwoertink, line 99: This stops the parser from parsing `-h` as the help for the `lucky` command. But we don't want to stop the parser when running internal commands like `lucky tasks -h` or `lucky dev -h`.

jwoertink:
I've been using this locally with my app, and it's felt good. Gonna merge this in!

PR #826, opened 2023-12-14
Snapshot /fixtures to avoid too long of a PR for feat/lucky-template-specs

mdwagner:
This is most of the changes from `feat/lucky-template-specs` branch, just being added prior so the actual PR can address what it needs to.


mdwagner:
gonna merge this, as its just static assets

PR #851, opened 2024-02-25
Chore: version bump to v1.1.1

mdwagner:


mdwagner:
ack: security spec will need to be looked into another time

PR #859, opened 2024-04-13
Updating generated CI with latest version

### LINKED ISSUE #817: Weak cache key for Crystal cache in generated CI workflow
The current cache key is often causing trouble when pushing changes to Github. It depends on any changes in the `shard.lock` file:

```
key: ${{ runner.os }}-crystal-${{ hashFiles('**/shard.lock') }}
```

If, for example, something is changed in a migration that has been committed before, the CI is going to fail. Unless a new dependency has been installed or the `shard.lock` file has been changed in some other way.

I'm not yet sure what a better approach would be; I'm just leaving this here as I've had issues with this across multiple projects.

**UPDATE**:
Looking at the code, it's probably because lucky_tasks is in the cache.

jwoertink:
Fixes #817

This updates several versions in the generated CI including a later version of Postgres, action updates, and also removes the tasks binary cache. Since the migrations run through tasks, if your tasks are cached and you make an update to migrations, they may not run properly.

Also, the Sectester JS deal isn't necessary anymore, so we don't even need to generate that.


jwoertink:
@mdwagner How does this work? I see there's some expected template in fixtures. Do I have to update the fixture to match what I'm looking for, or is there a command that does it? Also, I'm updating the minimum Crystal version. I see they all say 1.6.2, where does that get defined?

mdwagner:
There is a command to update a snapshot, it's in the Earthfile (update-snapshot), but if you don't want to use that, just look at what it runs.

mdwagner:
And I think the shard file generator spec defines a particular version.

PR #864, opened 2024-05-08
Fix: Release token

mdwagner:
I think this should work, but haven't tested it. Could try it next release.


jwoertink (approved):
Sweet! So this should fix that issue I had before?

mdwagner:
I think so. This was just one of the ways to accomplish this. We could have used a PAT, but then we'd have to cycle the token. So this was a way to keep it working unattended.

PR #867, opened 2024-09-08
Adding in setup scripts for Windows powershell users

jwoertink:
Fixes https://github.com/luckyframework/lucky_cli/issues/872

This adds in the next step to Windows compatibility. When you generate a Lucky app on Windows, it will now generate PowerShell scripts for you to run making the flow feel similar to mac/linux.

Full disclaimer... I used AI to translate bash to powershell then followed it up with lots of stackoverflow posts :joy:  I'll try to run this on an actual Windows machine to verify, but I'm just getting the PR up so we can poke around with it.

### DIFF
--- a/fixtures/src_template/expected/script/helpers/function_helpers.ps1
+++ b/fixtures/src_template/expected/script/helpers/function_helpers.ps1
@@ -1,0 +1,42 @@
+# This file contains a set of functions used as helpers
+# for various tasks. Read the examples for each one for
+# more information. Feel free to put any additional helper
+# functions you may need for your app
+
+
+# Returns true if the command $Command is not found
+# example:
+#   if (CommandNotFound "yarn") {
+#     echo "no yarn"
+#   }
+function CommandNotFound($Command) {
+  try {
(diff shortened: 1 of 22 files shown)


jwoertink:
```
'D:\\a\\lucky_cli\\lucky_cli\\spec/../fixtures\\src_template\\expected\\script/setup.ps1': The system cannot find the file specified.
```

My guess is the fixtures aren't setup to handle switching formats since they are statically generated. The weird thing though is why do the slashes go both directions? It says `script/setup.ps1`, but I'm using `Path.new` which should be setup for Windows properly...

jwoertink:
Well... I now have new errors which is good, but I still don't quite understand them.

jwoertink:
I had a thought on this... what if instead of using bash/powershell, it was just a Crystal bootstrap? Would be nice if it could just be Ruby or something to make it a lot faster, but that means a new language requirement. We at least know Crystal will exist when this runs, and things like `yarn` can just run through `Process`.

jwoertink:
I'm closing this in favor of #875

PR #871, opened 2024-10-06
Add option to run non-compiled crystal scripts from bin

### LINKED ISSUE #869: Add new CLI option to execute raw crystal files in bin
Related to [this comment](https://github.com/luckyframework/lucky/issues/1746#issuecomment-2381027036), I gave it a shot locally, and I think this might work a lot better...

What needs to happen here is 2 things

* Add an option when running something like `lucky gen.migration` if there's a `lucky.gen.migration` file, execute that as binary, but if there's a `lucky.gen.migration.cr` file then just run that through crystal

jwoertink:
Fixes #869

If you have a non-compiled crystal script in `bin` like `bin/lucky.some_script.cr`, running `lucky some_script` will now execute that script.

### DIFF
--- a/spec/integration/lucky_cli_spec.cr
+++ b/spec/integration/lucky_cli_spec.cr
@@ -9,8 +9,19 @@ describe "Lucky CLI", tags: "integration" do
         shell: true,
         output: io
       )
-      status.exit_status.should eq(0)
+      status.exit_code.should eq(0)
       io.to_s.should eq("Hello World!\n")
+    end
+
+    it "runs non-compiled tasks" do
+      io = IO::Memory.new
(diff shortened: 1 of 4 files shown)


spec/integration/lucky_cli_spec.cr
```diff
@@ -9,10 +9,21 @@ describe "Lucky CLI", tags: "integration" do
         shell: true,
         output: io
       )
-      status.exit_status.should eq(0)
+      status.exit_code.should eq(0)
```
jwoertink, line 12: These changes are related to https://github.com/crystal-lang/crystal/issues/8381

PR #884, opened 2025-06-06
Adding for_collection to the BaseSerializer.

russ:
This for_collection override takes a Lucky::Paginator argument as an option and changes the output to have extra pagination data added in.


russ:
Oops. That would make sense wouldn't it. lol

jwoertink:
For this, you may need to run `earthly +update-snapshot` locally. I think that should update the fixtures and set all of this

jwoertink (approved):

PR #924, opened 2026-04-03
Moving from Vite default to Bun default

### LINKED ISSUE #893: Default generated apps to using Vite instead of Mix
Related: https://github.com/luckyframework/lucky/issues/1980

Vite is much faster than Mix, and with Bun about to be default instead of yarn, we might as well do the swap here too.

One thing to consider is that we need to maintain backwards compatibility so when someone using mix updates Lucky, their app doesn't just bork all over.

jwoertink:
Closes https://github.com/luckyframework/lucky_cli/pull/917
Closes https://github.com/luckyframework/lucky_cli/issues/916
Closes https://github.com/luckyframework/lucky_cli/issues/913
Closes https://github.com/luckyframework/lucky_cli/issues/893

Instead of default all new apps to Vite, @wout has put in some serious effort to just making Bun work out of the box. ref https://github.com/luckyframework/lucky/issues/1931 

This PR sets up all new browser apps to just use Bun out of the box. This requires no additional dependencies, it's 100x faster at bundling assets, supports typescript, and static assets with cache busting, and no special config is needed to make things "just work". 

Now, Webpack and Vite will still be usable (at least for a while), so existing apps won't have to switch over, but this gives new apps a huge benefit in simplicity.


jwoertink:
There is possibly one strange bug I got locally while testing... When setting up a fresh app on my local machine, it all worked perfect, but when booting the app in Docker, I was getting this infinite page reload happening caused by the `bun_reload_connect_tag`. Maybe I have something wrong in Docker, or it's a legit bug... I'm not sure.

src/browser_app_skeleton/src/components/shared/layout_head.cr.ecr
```diff
@@ -11,8 +11,12 @@ class Shared::LayoutHead < BaseComponent
       responsive_meta_tag
 
       # Development helper used with the `lucky watch` command.
-      # Reloads the browser when files are updated.
+      # Reloads the browser when Crystal files are updated.
       live_reload_connect_tag if LuckyEnv.development?
```
jwoertink, line 15: @wout is this still needed? It sort of looked like the `bun_reload_connect_tag` also handles crystal files?

src/generators/web.cr
```diff
@@ -103,9 +103,8 @@ class LuckyCli::Generators::Web
       ignore_list.concat([
         "/public/js",
         "/public/css",
-        "/public/.vite",
         "/public/assets",
-        "/public/manifest.dev.json",
+        "/public/bun-manifest.json",
```
jwoertink: Is there a way to just tell .gitignore to ignore any .json file that has `manifest` in the name? 😂

src/browser_app_skeleton/src/components/shared/layout_head.cr.ecr
wout: It's something I've been thinking about too. It feels like Bun could handle Crystal changes too, since it's also a WebSocket server. In the end I decided it's best to keep things separate. Bun handles the frontend with `bun_reload_connect_tag`, Lucky the backend with `live_reload_connect_tag`. Because if someone decides not to use Bun, they would lose live reloading for Lucky.

src/generators/web.cr
wout: With a glob pattern?
`/public/*manifest*.json`

src/browser_app_skeleton/src/components/shared/layout_head.cr.ecr
jwoertink: Fair. I'm ok with keeping both. :+1:

